PR #5, opened 2022-02-07
added macro to set aliases for functions to compute different parameters

mcosarinsky:
Included the possibility to set names aliases for methods used to compute different parameters.


boriskaus:
thanks, great work!

PR #6, opened 2022-03-21
New methods for dislocation creep laws and fix test_DislocationCreep.jl

albert-de-montserrat:
Currently `computeCreepLaw_EpsII()` and `computeCreepLaw_TauII()` take the argument `p::CreepLawVariables`. To use these routines with [ParallelStencil.jl](https://github.com/omlins/ParallelStencil.jl) it's better to have a bit less abstraction. This PR adds a couple of methods `computeCreepLaw_EpsII()` and `computeCreepLaw_TauII()` where the arguments `P,T,f` are `Real` numbers instead of being inside `CreepLawVariables`. The corresponding test file is also fixed.


boriskaus:
yes, I fully agree. This has been on my to-do list for some time. 
Thanks a lot!

PR #7, opened 2022-03-23
Add functors to viscosity laws

albert-de-montserrat:
I have been doing a bit of more work trying to make the viscosity laws in `GeoParams.jl` compatible with `JustRelax.jl` and `ParallelStencil.jl` . I have come up with a simple solution that should work for single and multiple phases. The idea is just to  add functors to the viscous laws `structs` (e.g. [here](https://github.com/albert-de-montserrat/GeoParams.jl/blob/365a68deedb5bd0fe5188bf460dfdd6e9e2bfa4a/src/CreepLaw/CreepLaw.jl#L73), [here](https://github.com/albert-de-montserrat/GeoParams.jl/blob/365a68deedb5bd0fe5188bf460dfdd6e9e2bfa4a/src/CreepLaw/CreepLaw.jl#L75), and [here](https://github.com/albert-de-montserrat/GeoParams.jl/blob/365a68deedb5bd0fe5188bf460dfdd6e9e2bfa4a/src/CreepLaw/CreepLaw.jl#L80)). This approach should be extensible to density or any other field.

For a single phase the usage  for dislocation creep would be something like this:
```Julia
using GeoParams
using ParallelStencil
using ParallelStencil.FiniteDifferences3D
@init_parallel_stencil(Threads, Float64, 3)

ni = (64,64,64) # grid size
EpsII,T,P,f=rand(ni...),rand(ni...),rand(ni...),rand(ni...)
η = DislocationCreep() # dislocation creep parameters for a single phase

# nodal viscosity
ηdis_i = η(EpsII[1,1,1],T[1,1,1],P[1,1,1],f[1,1,1])

# full viscosity field using broadcasting
ηdis = η.(EpsII,T,P,f)

# parallel stencil kernels
@parallel_indices (i,j,k) function foo!(A, η, EpsII, T, P, f)
    A[i,j,k] = η(EpsII[i,j,k],T[i,j,k],P[i,j,k],f[i,j,k])
    return nothing
end

A = zeros(ni...)
@parallel (1:ni[1],1:ni[1],1:ni[1]) foo!(A, η, EpsII, T, P, f)
```

and for multiple phases:

```Julia
nphase = 5 # number of phases
phase = rand(1:nphase, ni...) # nodal phase

# option (1): ntuple
ηtuple = ntuple(_ -> DislocationCreep(), Val(nphase))
# option (2): vector
ηvect = [DislocationCreep() for _ in 1:nphase]

# nodal viscosity
ηtuple_i = ηtuple[1](EpsII[1,1,1],T[1,1,1],P[1,1,1],f[1,1,1])
ηvect_i = ηvect[1](EpsII[1,1,1],T[1,1,1],P[1,1,1],f[1,1,1])

# full viscosity field using broadcasting
ηdis_tuple = ηtuple[1].(EpsII,T,P,f)
ηdis_vect = ηvect[1].(EpsII,T,P,f)

# parallel stencil kernels
@parallel_indices (i,j,k) function bar!(A, η, phase, T, P, f)
    A[i,j,k] = η(phase[i,j,k], EpsII[i,j,k], T[i,j,k], P[i,j,k], f[i,j,k])
    # A[i,j,k] = η[phase[i,j,k]](EpsII[i,j,k], T[i,j,k], P[i,j,k], f[i,j,k]) # alternative syntax
    return nothing
end

@parallel (1:ni[1],1:ni[1],1:ni[1]) bar!(A, ηtuple, phase, T, P, f)
@parallel (1:ni[1],1:ni[1],1:ni[1]) bar!(A, ηvect, phase, T, P, f)
```

And then for practical reasons one could dump the grid viscosity field and all the different laws in a single `struct` :
```Julia
struct Viscosity{T1,T2}
    val::T1
    params::T2
end

# parallel stencil kernels
@parallel_indices (i,j,k) function bar2!(visc, phase, T, P, f)
    visc.val[i,j,k] = visc.params(phase[i,j,k], EpsII[i,j,k], T[i,j,k], P[i,j,k], f[i,j,k])
    return nothing
end

visc = Viscosity(rand(ni...), ηvect)
@parallel (1:ni[1],1:ni[1],1:ni[1]) bar2!(visc, phase, T, P, f)
```

This has been implemented for `LinearViscous`, `PowerlawViscous`, and `DislocationCreep`. @boriskaus is there any modification/additional feature that would be useful for your cases?

## Performance

I have also done some performance tests and while in CPUs there is not much difference between a `NTuple` or a `Vector`  of phases (apart from a fairly higher compile time for the first) ,  `Vector` performs much better with CUDA (and for some reason CUDA crashes with tuples of length between 40 and 80):

### Dislocation creep CPU
[perf_cpu_dislocation.pdf](https://github.com/JuliaGeodynamics/GeoParams.jl/files/8333976/perf_cpu_dislocation.pdf)

### Dislocation creep CUDA
[perf_cuda_dislocation.pdf](https://github.com/JuliaGeodynamics/GeoParams.jl/files/8333980/perf_cuda_dislocation.pdf)

### Power law creep CPU
[perf_cpu_powerlaw.pdf](https://github.com/JuliaGeodynamics/GeoParams.jl/files/8333983/perf_cpu_powerlaw.pdf)

### Power law creep CUDA
[perf_cuda_powerlaw.pdf](https://github.com/JuliaGeodynamics/GeoParams.jl/files/8333984/perf_cuda_powerlaw.pdf)

### DIFF
--- a/src/CreepLaw/CreepLaw.jl
+++ b/src/CreepLaw/CreepLaw.jl
@@ -64,11 +64,23 @@ or
 
 where ``\\eta_0`` is the reference viscosity [Pa*s] at reference strain rate ``\\dot{\\varepsilon}_0``[1/s], and ``n`` the power law exponent []. 
 """
-@with_kw_noshow struct LinearViscous{T,U} <: AbstractCreepLaw{T}    
+@with_kw_noshow struct LinearViscous{T,U} <: AbstractCreepLaw{T}
     η::GeoUnit{T,U}     =   1e20Pa*s                # viscosity
     η_val::T            =   1.0
 end
 LinearViscous(args...) = LinearViscous(convert(GeoUnit,args[1]), args[2])
+
+(η::LinearViscous)(args...) = η.η.val
+
+function (η::NTuple{N, LinearViscous})(I::Int64, args...) where N 
+    @assert I ≤ N  
+    return η[I]()
+end
+
+function (η::AbstractVector{LinearViscous{T1,T2}})(I::Int64, args...) where {T1,T2}
+    @assert I ≤ length(η)
+    return η[I]()
+end
 
 function param_info(s::LinearViscous) # info about the struct
     return MaterialParamsInfo(Equation = L"\tau_{ij} = 2 \eta  \dot{\varepsilon}_{ij}")
@@ -110,6 +122,27 @@ where ``\\eta`` is the effective viscosity [Pa*s].
     ε0::GeoUnit{T,U3}     =  1e-15*1/s;          # reference strainrate
 end
 PowerlawViscous(a...) = PowerlawViscous(convert.(GeoUnit,a)...)
+
+function (η::PowerlawViscous)(εII) 
+    # unpack
+    η0 = η.η0.val
+    ε0 = η.ε0.val
+    n = η.n.val
+    # stress
+    τII = η0 * (εII/ε0)
+    # return viscosity
+    return τII^n/εII
+end
+
+function (η::NTuple{N, PowerlawViscous})(I::Int64, εII) where N 
+    @assert I ≤ N  
+    return η[I](εII)
+end
+
+function (η::AbstractVector{PowerlawViscous})(I::Int64, εII)
+    @assert I ≤ length(η)
+    return η[I](εII)
+end
 
 # Print info 
 function show(io::IO, g::PowerlawViscous)  
--- a/src/CreepLaw/CreepLaw.jl
+++ b/src/CreepLaw/CreepLaw.jl
@@ -139,7 +139,7 @@ function (η::NTuple{N, PowerlawViscous})(I::Int64, εII) where N
     return η[I](εII)
 end
 
-function (η::AbstractVector{PowerlawViscous})(I::Int64, εII)
+function (η::AbstractVector{PowerlawViscous{T,U1,U2,U3}})(I::Int64, εII) where {T,U1,U2,U3}
     @assert I ≤ length(η)
     return η[I](εII)
 end
--- a/src/CreepLaw/DislocationCreep.jl
+++ b/src/CreepLaw/DislocationCreep.jl
@@ -54,6 +54,21 @@ DislocationCreep: n=3, r=0.0, A=1.5 MPa^-3 s^-1, E=476.0 kJ mol^-1, V=6.0e-6 m^3
 end
 DislocationCreep(args...) = DislocationCreep(NTuple{length(args[1]), Char}(collect.(args[1])), convert.(GeoUnit,args[2:end-1])..., args[end])
 
+# Compute dislocation creep viscosity
+function (η::DislocationCreep)(EpsII, P::Real, T::Real, f::Real) 
+    return 0.5*EpsII*(1/computeCreepLaw_TauII(EpsII, η, P, T, f))
+end
+
+function (η::NTuple{N, DislocationCreep})(I::Int64, args...) where N 
+    @assert I ≤ N  
+    return η[I](args...)
+end
+
+function (η::AbstractVector{DislocationCreep{T,N,U1,U2,U3,U4,U5}})(I::Int64, args...) where {T,N,U1,U2,U3,U4,U5}
+    @assert I ≤ length(η)
+    return η[I](args...)
+end
+
 function param_info(s::DislocationCreep)
     name = String(collect(s.Name))
     eq = L"\tau_{ij} = 2 \eta  \dot{\varepsilon}_{ij}"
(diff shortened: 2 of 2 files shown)


boriskaus:
Thanks! 
We did a bit of work along these lines already, but it was all implemented for Density only (not yet for creep laws).
I haven't been able to fully go through your approach but a few things that come to mind:
1. Is your approach fully non-allocating (crucial for PT approaches & took us a long time to do for Density)?
2. Will it work if we extend it to combined creep laws (say diffusion & dislocation), or even if we add elasticity/plasticity?
3. The overall idea of `GeoParams`  is to only define a single struct `SetMaterialParams` at the beginning of the simulation such that we can easily change the creep law from say, linear to power law without having to change the actual computational routines/solver itself. Is that compatible with this?
4. There are essentially two ways of defining the phases at each point: 
  - In the first approach, which I believe you follow here, every point can have a single (integer) phase and that phase is used for computation. 
  - The second approach, which we actually use in LaMEM, defines the fraction of every phase at every point (which sums to 1). That actually allows having somewhat smoother transitions which is useful for solver convergence.  
5. Will it be possible to directly call routines from GeoParam in PS codes in this way, without having to explicitly define such routines in the user code? Samuel is working on allowing vectors within PS Data arrays, which could come handy to define the phase ratio @ every point.

albert-de-montserrat:
1. Yes this approach is allocation free as long as everything remains type-stable
```Julia
@benchmark @parallel (1:ni[1],1:ni[1],1:ni[1]) bar!($A, $ηvect,  $phase, $EpsII, $T, $P, $f)
BenchmarkTools.Trial: 3061 samples with 1 evaluation.
 Range (min … max):  1.504 ms …   2.413 ms  ┊ GC (min … max): 0.00% … 0.00%     
 Time  (median):     1.554 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.630 ms ± 125.555 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%     

    ▃█▂
  ▁▃███▄▃▂▂▂▂▂▂▂▂▂▁▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▃▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁ ▂
  1.5 ms          Histogram: frequency by time        1.92 ms <

 Memory estimate: 6.70 KiB, allocs estimate: 60.

```
2.  What I had in mind (for now) was to do something like this in `JustRelax`
```Julia
struct myViscosity{T1,T2,T3,T4}
    val::T1
    dif_creep::T2 # from GeoParams
    dis_creep::T3 # from GeoParams
    elasticity::T4 # from GeoParams
end

function (η::Viscosity)(args1, args2, args3) # this could probably be a @generated function
    return 1/(1/η.dif_creep + 1/η.dis_creep +1/η.elasticity )
end
```
but probably you don't want something like this inside `Geoparams`. The other options would be:

- Define several `structs` in `GeoParams` with different combination of creep laws, but this requires manual work and mantaining more lines of code.
- Programatically build custom creep laws  types and methods with metaprogramming. This would be my preferred approach, the main disadvantage I can think at this moment is that this custom creep type does not exist at compile/parse time, and hence it can't be used to dispatch. I will think more about this in the next days.

3. Indeed, this is what I'm trying to achieve. Ideally I would like to have a generic `compute_viscosity()` function  in `JustRelax` with the lowest possible number of methods that works with and without `GeoParams` and is able to  use any creep law defined at the `main()` top level.

4. For simplicity the code above uses the phase=integer approach, but it should be straightforward to implement a weigthed sum as in LAMEM. 

5. Yes, right now it works with `@parallel_indices` but it should be possible to do something like
```Julia
@parallel function foo!(A, η, phase, EpsII, T, P, f)
    @all(A) = η(@all(phase), @all(EpsII), @all(T), @all(P), @all(f))
    return nothing
end
```
where `η` is a `GeoParams` creep law. For the phase ratio at grid points I guess it depends. Does the phase weight need to be explicitely stored? Can't it just be computed and used on the fly?

boriskaus:
1. Thanks for your explanation. It still states 60 allocs (?)

2. The idea is indeed to have a `compute_viscosity()` routine within `GeoParams`, rather than putting it in a user routines. This is because it will make future codes more comparable and future (new) creep laws implemented in `GeoParams` can directly be used in a range of other codes. This would indeed require some work, ideally to be done automatically. Note that the example you give does a harmonic averaging of viscosity, which is one way to do it but is actually not entirely correct if dislocation creep and elasticity are combined (which would require local iterations). We haven't thought very hard yet on how this can be achieved with creep laws that require different amounts of input parameters (not only P,T, but perhaps also H2O content or grainsize)

3. See above

4. Ok, would be good to have this option directly in `GeoParams` 

5. Yes that would do (or could be implemented in `GeoParam` & made available in case `ParallelStencil` is loaded. Phase weight is usually computed from markers which is a relatively slow process. Doing this once/timestep is fine but doing it every iteration during a pseudo transient simulations will likely really slow down the code. That's why storing it is likely necessary.

albert-de-montserrat:
1. Those are allocations from `Threads.@threads`, it always allocates few kbs. Nothing to worry about.
2. Great, I will look more into automatically doing it then. I'm not sure I follow the issue of different amount of parameters, you mean dealing with `CreepLaw1(a)` and `CreepLaw2(a,b)` ?
3.  
4. Sure.
5. I see, then the approach of Sam would be the way to go. It should combine well with what I'm trying to build.

Let's leave this PR open for a bit, I'll do some more work on this in the next few days.

boriskaus:
1. cool

2. correct

PR #8, opened 2022-04-09
Rework of all the compute_density methods

albert-de-montserrat:
**WARNING**: this PR contains code-breaking changes
`compute_density()` and `compute_density!()` have been reworked to work with an arbitrary number of input parameters. The argument corresponding to the density equation parameters is now a `NamedTuple` (i.e. `args=(;P=rand(20, 20)` or `args=(P=rand(20, 20), T = rand(20, 20)`; see `test/tes_Density.jl` for syntax examples).

Example of new usage to compute density with/without `ParallelStencil.jl`
```Julia
using GeoParams
using ParallelStencil
using ParallelStencil.FiniteDifferences2D
@init_parallel_stencil(Threads, Float64, 2)

# with ParallelStencil
@parallel_indices (i, j) function foo(rho, MatParam, Phases, args)
    rho[i,j] = compute_density(MatParam, Phases[i,j], ntuple_idx(args, i, j))
    return 
end

# standard Julia
function bar(rho, MatParam, Phases, args)
    for i in eachindex(rho)
        rho[i] = compute_density(MatParam, Phases[i], ntuple_idx(args, i))
    end
end

MatParam1 = Vector{AbstractMaterialParamsStruct}(undef, 4)
MatParam1[1] = SetMaterialParams(Name="Crust", Phase=0,
                            CreepLaws= (PowerlawViscous(), LinearViscous(η=1e23Pas)),
                            Density   = ConstantDensity(ρ=2900kg/m^3))
MatParam1[2] = SetMaterialParams(Name="Lower Crust", Phase=1,
                            CreepLaws= (PowerlawViscous(n=5.), LinearViscous(η=1e21Pas)),
                            Density  = Compressible_Density(ρ0=3000kg/m^3))
MatParam1[3] = SetMaterialParams(Name="Lower Crust", Phase=2,
                            CreepLaws= LinearViscous(η=1e21Pas),
                            Density  = ConstantDensity())
MatParam1[4] = SetMaterialParams(Name="Lower Crust", Phase=3,
                            CreepLaws= LinearViscous(η=1e21Pas),
                            Density  = ConstantDensity())
MatParam = Tuple(MatParam1)

Phases              = ones(Int64,400,400)*0;
Phases[:,20:end]   .= 1
Phases[:,200:end]  .= 2
Phases[:,300:end]  .= 3

rho     = zeros(size(Phases))
T       =  ones(size(Phases))
P       =  ones(size(Phases))*10
args = (P=P, T=T)

@parallel (1:size(Phases, 1), 1:size(Phases, 2)) foo(rho, MatParam, Phases, args) # ParallelStencil
bar(rho, MatParam, Phases, args) # standard Julia
```
@boriskaus It would be great if before merging you could test it out with your codes to see if it works fine or if there are any performance issues  (I have observed something like +20/40% speed in most of my tests).


albert-de-montserrat:
Will need some extra work, didnt realise this breaks all the other `compute_something()` functions

boriskaus:
That’s clear that more tests fail but not a big deal as this is an extremely useful and necessary PR. I will give it a go with some real world codes but this already looks very promising.

boriskaus:
On my machine, the new Julia version is actually faster than the existing routines:

**Previous:**
400x400:
```julia
julia> @btime compute_density!($rho, $MatParam, $Phases, $P,$T);
  929.500 μs (0 allocations: 0 bytes)
```
4000x4000:
```julia
julia> @btime compute_density!(rho, MatParam, Phases, P,T);
  104.833 ms (0 allocations: 0 bytes)
```

**This PR:**
400x400:
```julia
julia> @btime bar($rho, $MatParam, $Phases, $args) # standard Julia
  666.567 μs (0 allocations: 0 bytes)
```
4000x4000:
```julia
julia> @btime bar($rho, $MatParam, $Phases, $args) # standard Julia
  45.283 ms (0 allocations: 0 bytes)
```

One difference, though, is that `compute_density!` previously directly worked on arrays, whereas you here have to write your own loops over the arrays. From a user point of view I think we should provide such routines (perhaps by adding a multiple dispatch in case Phases is an array?). Similarly in case we use `ParallelStencil` (which was not yet implemented in the previous version, perhaps to be done with using [Requires.jl](https://github.com/JuliaPackaging/Requires.jl)?).

albert-de-montserrat:
Cool! That's a much better performance boost than I expected :)  That `bar()` function is just a toy example for a one-to-one comparison with the syntax of a `ParallelStencil` kernel. All the pre-existing density methods still exist using the new code under the hood, for example [here](https://github.com/albert-de-montserrat/GeoParams.jl/blob/2d221ff8a639d3e983574b41ce7137193e709662/test/test_Density.jl#L166) for the in-place function using an integer array for the phase, and [here](https://github.com/albert-de-montserrat/GeoParams.jl/blob/2d221ff8a639d3e983574b41ce7137193e709662/test/test_Density.jl#L193) using phase ratios. I covered all the methods in `test_Density.jl` but if there's any missing functionality let me know and I can add it.

As for providing `ParallelStencil` kernels, I'm not sure what is the best approach. The problem with the current status of PS is that before parsing custom kernel with `@parallel` or `@parallel_indices` one needs to run `@init_parallel_stencil` to set up the device and dimension. Meaning that `Requires` will load our kernels as soon as `PS` is loaded, and crash. What about building a `GeoParams.ParallelStencilKernels` submodule that the user can optionally load after setting up `PS` ?

Btw I realised that the densities of type `PhaseDiagram_LookupTable` work, but are not type-stable. I didn't look in detail at what is the problem, but we should fix this before running it on GPUs.

boriskaus:
Two additional use case are the following:

**1. Loop**
Typically we would want to use this within a timestep or iteration loop in which `P`,`T` changes. 
If I define the following:
```julia
function loop(rho,P,T,MatParam,Phases)
    for iter=1:10
        P .= rand(size(P))
        T .= rand(size(T))
        args = (P=P, T=T);
        bar(rho, MatParam, Phases, args) 
    end
    nothing
end
```
it indeed does not allocate: 
```julia
julia> @btime loop($P,$T,$rho,$MatParam, $Phases);
  667.065 ms (0 allocations: 0 bytes)
```

**2. Additive functions**
This is not the case for density, but for creep laws we often have several functions in parallel (say a dislocation and a diffusion creep law) and make a sum of the contribution of each (if using a strain rate decomposition). 
It is important to test that this is still non-allocating. As density only allows one input parameter [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/49ec5e4793a43805ad3fab2f6c2dd2e134a04886/src/MaterialParameters.jl#L196), we would have to temporarily deactivate this.

boriskaus:
> Btw I realised that the densities of type `PhaseDiagram_LookupTable` work, but are not type-stable. The problem is all the `::Any` annotations [here](https://github.com/albert-de-montserrat/GeoParams.jl/blob/2d221ff8a639d3e983574b41ce7137193e709662/src/PhaseDiagrams/PhaseDiagrams.jl#L18). Will all these paremeters always have random and different types?

No, most of them are `Float64`, so the structure can be made type stable. 
Moving the whole struct to the GPU is a different problem as we have strings as well (which are not actually required for computations on the GPU, but are mostly for info about the phase diagram). Moreover, we use the [Interpolations.jl](https://github.com/JuliaMath/Interpolations.jl) package to do the linear interpolation within the phase diagram. This does currently not work on the GPU (but I guess we can write a custom interpolation routine if this becomes a bottleneck).

albert-de-montserrat:
> 



> Two additional use case are the following:
> 
> **1. Loop** Typically we would want to use this within a timestep or iteration loop in which `P`,`T` changes. If I define the following:
> 
> ```julia
> function loop(rho,P,T,MatParam,Phases)
>     for iter=1:10
>         P .= rand(size(P))
>         T .= rand(size(T))
>         args = (P=P, T=T);
>         bar(rho, MatParam, Phases, args) 
>     end
>     nothing
> end
> ```
> 
> it indeed does not allocate:
> 
> ```julia
> julia> @btime loop($P,$T,$rho,$MatParam, $Phases);
>   667.065 ms (0 allocations: 0 bytes)
> ```
 here you can already do 
```Julia
function loop(rho,P,T,MatParam,Phases)
    for iter=1:10
        P .= rand(size(P))
        T .= rand(size(T))
        args = (P=P, T=T);
        compute_density!(rho, MatParam, Phases, args) 
    end
    nothing
end
```
> **2. Additive functions** This is not the case for density, but for creep laws we often have several functions in parallel (say a dislocation and a diffusion creep law) and make a sum of the contribution of each (if using a strain rate decomposition). It is important to test that this is still non-allocating. As density only allows one input parameter [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/49ec5e4793a43805ad3fab2f6c2dd2e134a04886/src/MaterialParameters.jl#L196), we would have to temporarily deactivate this.

Yep, I've been adapting the energy functions first. I'll move to viscosity next.

boriskaus:
This is a great PR, which makes it easier to deal with computational methods that require an arbitrary number of input fields. 

For now it works for all thermal and density parameters, but not yet for creep laws. Yet, we are merging now, so it can already be used in a real code, and such that it is also clear how new methods will have to be added.

PR #14, opened 2022-05-04
Elasticity & Plasticity - 2 

boriskaus:
This slightly updates the elasticity & plasticity implementation (simplifying it and adding fluid pressure). It also adds the comments of Albert on the previous PR (which was merged too quickly)

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -15,6 +15,7 @@ makedocs(
             "Material Parameters" =>  "man/materialparameters.md",
             "Density" =>  "man/density.md",
             "Creep laws" =>  "man/creeplaws.md",
+            "Elasticity" =>  "man/elasticity.md",
             "Plasticity" =>  "man/plasticity.md",
             "Heat Capacity" =>  "man/heatcapacity.md",
             "Conductivity" =>  "man/conductivity.md",
--- a/docs/src/man/elasticity.md
+++ b/docs/src/man/elasticity.md
@@ -1,0 +1,16 @@
+# Elasticity 
+
+Elasticity is, in geodynamics, often used in combination with viscous and plastic rheology.
+
+# Implemented laws
+We provide the following elastic constitutive relationships:
+```@docs
+GeoParams.MaterialParameters.Elasticity.ConstantElasticity
+```
+
+# Computational routines 
+We can compute the elastic strainrate with:
+```@docs
+GeoParams.MaterialParameters.Elasticity.compute_elastic_shear_strainrate
+GeoParams.MaterialParameters.Elasticity.compute_elastic_shear_strainrate!
+```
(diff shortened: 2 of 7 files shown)


boriskaus:
If we have elasticity, we need to pass the timestep `dt` to the computational routine. As it is just a scalar, we would ideally pass it as a scalar:
```Julia
τII      =  ones(size(Phases))*20e6;
τII_old  =  ones(size(Phases))*15e6;
args = (τII=τII, τII_old=τII_old, dt=1e6);
```
That works when the Phases are defined as an integer array:
```Julia
compute_elastic_shear_strainrate!(ε_el, MatParam, Phases, args) 
```
as done [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/584b1a9c441c0618a4b9d3b08dd256732f4ebe43/test/test_Elasticity.jl#L56).

Yet, it does not work when we provide a `PhaseRatio` array. Instead we have to provide an array with timesteps as done [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/584b1a9c441c0618a4b9d3b08dd256732f4ebe43/test/test_Elasticity.jl#L70)

Fixing this will require changing this routine 
https://github.com/JuliaGeodynamics/GeoParams.jl/blob/584b1a9c441c0618a4b9d3b08dd256732f4ebe43/src/Computations.jl#L73
.
@albert-de-montserrat: perhaps you have an idea how this can be implemented.

albert-de-montserrat:
> 

Adding/overloading methods of`getindex()` for scalar values seems to do the trick:
```julia
Base.getindex(val::Number, I::Vararg{Integer, N}) where N = val
Base.getindex(val::Number, I::Integer) = val
```
then 
```julia
n = 100;
Phases              = ones(Int64,n,n,n);
Phases[:,:,20:end] .= 2;
Phases[:,:,60:end] .= 2;

PhaseRatio  = zeros(n,n,n,3);
for i in CartesianIndices(Phases)
    iz = Phases[i]
    I = CartesianIndex(i,iz)
    PhaseRatio[I] = 1.0  
end

τII      =  ones(size(Phases))*20e6;
τII_old  =  ones(size(Phases))*15e6;
ε_el = zero(τII);
args = (τII=τII, τII_old=τII_old, dt=1e6);

I= CartesianIndices(ε_el)[2]
v = getindex.(values(args), Tuple(I)...) # (2.0e7, 1.5e7, 1.0e6)
```

If this has any weird side effects we can replace `getindex()` for a custom one.

boriskaus:
Impressive, that indeed does the trick! 
Here some timings:
```Julia
 args = (τII=τII, τII_old=τII_old, dt=dt_arr);
julia> @btime compute_elastic_shear_strainrate!($ε_el, $MatParam, $PhaseRatio, $args)
  7.496 ms (0 allocations: 0 bytes)
```
```Julia
julia> args1 = (τII=τII, τII_old=τII_old, dt=1e6);
julia> @btime compute_elastic_shear_strainrate!($ε_el, $MatParam, $PhaseRatio, $args1)
  6.233 ms (0 allocations: 0 bytes)
```
Slight saving (as expected).

PR #17, opened 2022-05-18
Seismic velocity correction for melt fraction

NicolasRiel:
The routine uses the reduction formulation of Clark et al., (2017) and is based on the equilibrium geometry model for the solid skeleton of Takei et al., 1998.


boriskaus (approved):
thanks a lot - great addition!

PR #18, opened 2022-05-19
Anelastic correction for S-wave velocity

NicolasRiel:
The routine uses the reduction formulation of karato (1993), using the quality factor formulation from Behn et al. (2009) to compute the anelastic contribution of S-wave velocities


boriskaus:
can you add the full references to the papers you used in the help text of the function?

NicolasRiel:
Added the references

PR #19, opened 2022-06-05
corrected zircon PD function

NicolasRiel:
The function now also returns the cumsum of PD zircon ages in [%]


boriskaus:
thanks!

PR #25, opened 2022-06-30
Added creep laws and a table function for output

Dr3ven:
Added two creep laws (Avery Island rock salt, Rock salt (Li & Urai)) as well as two functions which produces a LaTex code for a parameter table. Automatic references are still to be added to the table.


boriskaus:
please note that we are reworking the viscosity calculations within the `/viscosity` branch. If you add new creep laws, please make a PR to that branch.

boriskaus:
also, you are adding many different things within one PR. Would be great to split the rheology from other things (such as creating a table, which deserves a separate PR).

PR #27, opened 2022-07-04
update porosity correction for s-waves

boriskaus:
This incorporates the suggestion of @albert-de-montserrat

### DIFF
--- a/src/PhaseDiagrams/PhaseDiagrams.jl
+++ b/src/PhaseDiagrams/PhaseDiagrams.jl
@@ -32,6 +32,13 @@ struct PhaseDiagram_LookupTable <: AbstractPhaseDiagramsStruct
     Vs::Any
     VpVs::Any
     cpxFrac::Any
+    solid_Vp::Any
+    solid_Vs::Any
+    melt_bulkModulus::Any
+    solid_bulkModulus::Any
+    solid_shearModulus::Any
+    Vp_corrected::Any           # will hold Vs velocity corrected for pores, fluids, & melt 
+    Vs_corrected::Any
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/SeismicVelocity/SeismicVelocity.jl
```diff
@@ -243,43 +243,54 @@ function porosity_correction(
...
-        1.549 4.814 8.777 -0.290
-    ] 
+    ν = 0.25 # poisson ratio
 
-    bij = [
-        -0.3238 0.2341
-        -0.1819 0.5103
-    ]
+    # Tuples are a better option than standard arrays for matrices/vectors of known size at compile time
+    aij = (
```
albert-de-montserrat, line 256: The same  thing should be done in the function `melt_correction()`, which is currently using arrays.

src/PhaseDiagrams/PhaseDiagrams.jl
```diff
@@ -32,6 +32,13 @@ struct PhaseDiagram_LookupTable <: AbstractPhaseDiagramsStruct
     Vs::Any
     VpVs::Any
     cpxFrac::Any
+    solid_Vp::Any
```
albert-de-montserrat, line 35: This is `struct` is not parametric, which may result in type instabilities.  I would make it parametric with an inner constructor which promotes all the variables that are currently annotated as `Any` to the same type. Something like this:
```julia
struct PhaseDiagram_LookupTable{S,T} <: AbstractPhaseDiagramsStruct
    Type::S
    Name::S
    HeaderText::T
    rockRho::T
    meltRho::T
    etc::T
    function PhaseDiagram_LookupTable(Type, HeaderText, Name, rockRho, meltRho, etc)
        args = promote(HeaderText, rockRho, meltRho, etc)
        T = eltype(args)
        new{String, T}(Type, Name, args...)
    end
end
```

PR #30, opened 2022-08-18
Method to compute gravity given material phases

albert-de-montserrat:
This fixes the problem where a call like this `compute_gravity(MatParam[Phases[i,j]].Gravity[1])` is type-unstable. The new type-stable method can now be called as  `compute_gravity(MatParam, Phases[i,j])`.


boriskaus (approved):
great, thanks!

PR #31, opened 2022-09-02
Tensor algebra

albert-de-montserrat:
Add routines to locally compute  the second invariant of a tensor given in the form of a `Matrix`, `SMatrix`or `NTuple`. Also adds `StaticArrays` as a dependency


boriskaus:
What about adding `structs` as well, which hold 2D and 3D tensors? As discussed, I think it is important that they do not only have the components:
```Julia
struct tensor2D{T}
txx::T
txy::T
tyy::T
end
```
but also have the possibility to hold second invariants, because if we deal with staggered grids the second invariant should be computed as the sum of squares of the components at their respective location, before averaging this to the center or corner points (it can be demonstrated that averaging `Exy` to the center points gives a different results than averaging `sqrt(Exy^2)`, as plus & minus signs cancel.

albert-de-montserrat:
Mmh then it gets implementation-dependent, as it is different for FEM codes. May it be worth having a point-wise generic version, and a special one for staggered grids?

boriskaus:
That is correct - one could also supply a differently computed strainrate tensor for center and corner points in case of FD

albert-de-montserrat:
What about this for staggered grids:
```julia 
Aii = rand(), rand()
Axy = rand(), rand(), rand(), rand() # essentially (Axy[i, j], Axy[i+1, j+1], Axy[i+1, j], Axy[i, j+1])

second_invariant_staggered(Aii, Axy)
```
 
where `Aii` are the diagonal terms, and `Axy` are the components that need to be interpolated to the center of the cell.  The routine internally averages the square of the off-diagonal terms.

I think for now I do not want to include tensor `struct`s in GeoParams as the memory map for those may change for different codes.  So rather provide the generic point-wise kernel and let the user define their own wrapper around it.

boriskaus:
that would work, but please note that sometimes we need to average `Aii` from centers -> vertexes

boriskaus (approved):

PR #32, opened 2022-09-06
Viscosity

albert-de-montserrat:
Implementation of routines to compute rheologies for arbitrary Maxwell bodies.


boriskaus (approved):
Thanks - great job!

PR #33, opened 2022-09-08
type instability

albert-de-montserrat:
Fix type instability in [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/a81bf559b8ed6e9f42920210af3bb6c2cfbd6689/src/Computations.jl#L25) which led to a 40% performance drop.


boriskaus (approved):
Fantastic - thanks a lot!

PR #36, opened 2022-09-20
Fix composite rheology bugs and  improve performance of  compute_viscosity() by 60%

albert-de-montserrat:


boriskaus (approved):
Great catch!

PR #38, opened 2022-09-27
Precompute trigonometric functions

albert-de-montserrat:
Trigonometric functions are expensive and the `sind` and `cosd`  calls on the friction and dilation angles can be precomputed within the plasticity struct. This removes a lot of redundant compute time:

```julia
 a=DruckerPrager()

julia> @btime $a() # trigs not precomputed
  17.735 ns (0 allocations: 0 bytes)
-8.660254037844386e6

julia> @btime $a() # trigs precomputed
  2.000 ns (0 allocations: 0 bytes)
-8.660254037844386e6
```


boriskaus (approved):

boriskaus:
wow, that's indeed a big difference

PR #39, opened 2022-09-29
Add plastic multiplier (lambda) and plastic strain

albert-de-montserrat:


boriskaus (approved):
ok for DP plasticity, lambda is indeed analytically known. That is not generally the case, though, and for more complicated plasticity models we actually have to iterate for lambda

PR #40, opened 2022-09-30
Composite Rheologies

boriskaus:
This PR brings support for composite rheologies. 
In many geoscience applications, it is typical to have different rheologies active such as diffusion creep and dislocation creep, or viscosity-elasticity (or visco-elasto-viscoplasticity).
In general we can distinguish between two ways of putting together rheological elements:  

1. serial:
```julia
           --⟦▪̲̅▫̲̅▫̲̅▫̲̅----⟦▪̲̅▫̲̅▫̲̅▫̲̅----/\/\/--
ε_total =     ε1   +  ε2   +    ε3
τ_total =     τ1   =  τ2   =    τ3
```
2. parallel
```julia
|--⟦▪̲̅▫̲̅▫̲̅▫̲̅--|  ε1, τ1
|--⟦▪̲̅▫̲̅▫̲̅▫̲̅--|  ε2, τ2
|--/\/\/--|  ε3, τ3

ε_total = ε1 = ε2 = ε3
τ_total = τ1 + τ2 + τ3
```
These can be combined in arbitrary manners. 
In the serial case, the sum of the strain rates is equal to the applied one (ε_total) and the deviatoric stress of each of the elements is equal. In the parallel case, the sum of the stresses is equal to the total (τ_total) and the strain rate of each of the components is equal. 
If one of the elements has a nonlinear dependence of stress one needs to solve a nonlinear system of equations. If a serial element has one or several parallel elements as well the problem requires solving several nonlinear equations at the same time. 

With this PR, you can specify a serial element with the `CompositeRheology` structure and parallel elements with the `Parallel` struct:
```julia
julia> v1 = SetDiffusionCreep("Dry Anorthite | Rybacki et al. (2006)");
julia> v2 = SetDislocationCreep("Dry Anorthite | Rybacki et al. (2006)");
julia> v3 = LinearViscous();
julia> e1 = ConstantElasticity();          
julia> c1 = CompositeRheology(v1,v2, e1)
--⟦▪̲̅▫̲̅▫̲̅▫̲̅----⟦▪̲̅▫̲̅▫̲̅▫̲̅----/\/\/--
julia> p1 = Parallel(e1,v3,v2)
|--/\/\/--|
|--⟦▪̲̅▫̲̅▫̲̅▫̲̅--|
|--⟦▪̲̅▫̲̅▫̲̅▫̲̅--|
```
you can also combine them:
```julia
julia> c2 = CompositeRheology(v1,v3, Parallel(v1,v2))
--⟦▪̲̅▫̲̅▫̲̅▫̲̅----⟦▪̲̅▫̲̅▫̲̅▫̲̅--|--⟦▪̲̅▫̲̅▫̲̅▫̲̅--|
                 |--⟦▪̲̅▫̲̅▫̲̅▫̲̅--|
```
Once defined, it will do the nonlinear iterations and return the solution, for example:
```Julia
julia> εII=2e-15;
julia> args = (T=900.0, d=100e-6, τII_old=1e6);
julia> τII = compute_τII(c1, εII, args, verbose=true)
initial τII = 999107.3731234319
 iter 1 0.0008926270763892077
 iter 2 0.0
final τII = 1.0000000001999997e6
---
1.0000000001999997e6
julia> compute_τII(c2, εII, args, verbose=true)
τII guess = 399999.39370252617
 iter 1 3.0278098697184688e-6
 iter 2 1.7084983666410965e-12
---
399999.6965107696
```
By default, this will do Newton iterations using analytical derivatives. Yet, it is also possible to use automatic differentiation to compute the required derivatives, by calling:
 
```julia
julia> τII = compute_τII_AD(c2, εII, args)
399999.69651076966
```

We also provide routines that return strain rate for a given stress:
```Julia
julia> compute_εII(c2, 1e6, args)
5.000003794807172e-15
```
or effective viscosity for a given strain rate 
```Julia
julia> computeViscosity_εII(c2, εII, args)
9.999992412769239e19
```

We have a new  `CompositeRheology` in the `MaterialParams` structure:
```Julia
julia> MatParam_nd = SetMaterialParams(Name="Viscous Matrix", Phase=2,
                   Density   = ConstantDensity(),
                   CompositeRheology   = c5, CharDim=GEO_units())
Phase 2 : Viscous Matrix
        | [non-dimensional units]
        | 
        |-- Density           : Constant density: ρ=2.8999999999999996e-18 
        |-- Gravity           : Gravitational acceleration: g=9.81e20 
        |-- CompositeRheology : --⟦▪̲̅▫̲̅▫̲̅▫̲̅----⟦▪̲̅▫̲̅▫̲̅▫̲̅--|--⟦▪̲̅▫̲̅▫̲̅▫̲̅--|
                                                 |--⟦▪̲̅▫̲̅▫̲̅▫̲̅--|
```
As you see, non-dimensionalisation is taken care of.

*Limitations*
There are a number of things not yet supported. We decided to leave them out for now to not make the PR even larger.
- Plasticity is not yet tested
- There are no computational routines for arrays yet
- We do not dispatch the computations if providing a `MatParam` structure.
- We do not yet automatically construct a `CompositeRheology` structure if parameters are specified in `SetMaterialParams` as `Elasticity`, `CreepLaws` and `Plasticity`, instead of directly within the `CompositeRheology` structure. It would be handy to introduce an automatic translation of the most commonly used geodynamic mechanisms into a `CompositeRheology` struct, which is than used to dispatch calculations. 
- The computational routines are strictly for shear stresses/shear deformation and not yet for bulk/volumetric deformation
- Cases with multiple elastic elements would need multiple `args` as each elastic element should have its own stress. Similarly, one can imagine cases with multiple viscous creeplaws where each, for example, has its own grain size.


src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+@generated function fill_J_parallel!(J, r, x, c::CompositeRheology{T, N, Npar, is_par, τ_it, P_it, λ_it}, τ, args) where {T, N, Npar, is_par, τ_it, P_it, λ_it}
+    quote
+        Base.@_inline_meta
+        j = 1
+        Base.Cartesian.@nexprs $N i -> j = @inbounds _fill_J_parallel!(J, r, x, c.elements[i], τ, args, $(is_par)[i], j)
+        return nothing
+    end
+end
+
+function _fill_J_parallel!(J, r, x, elements, τ, args, is_par, j)
```
albert-de-montserrat, line 553: This is allocating because the compiler is failing to inline this function.  Adding the `@inline` macro here removes all the allocations.

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+        τ   = x[1]
+  
+        # Update part of jacobian related to serial elements
+        r[1]   = εII_total - compute_εII_elements(c,τ,args)
+        J[1,1] = dεII_dτII_elements(c,x[1],args);
+        
+        # Add contributions from || elements
+        fill_J_parallel!(J, r, x, c, τ, args)
+        
+        #=
```
albert-de-montserrat, line 507: These commented lines can be removed

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+    while ϵ > tol
+        iter += 1
+        #= 
+            Newton scheme -> τII = τII - f(τII)/dfdτII. 
+            Therefore,
+                f(τII) = εII - strain_rate_circuit(v, τII, args) = 0
+                dfdτII = - dεII_dτII(v, τII, args) 
+                τII -= f / dfdτII
+        =#
+        # f = τII - compute_τII(v, εII, args)
```
albert-de-montserrat, line 391: The three commented lines below can be removed

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+        Base.@_inline_meta
+        τII = zero(_T)
+        Base.Cartesian.@nexprs $N i ->
+            τII += compute_τII(v.elements[i], εII, args)
+    end
+end
+compute_τII_AD(v::Parallel{T,N}, εII::_T, args; tol=1e-6, verbose=false) where {T,N,_T} = compute_τII(v, εII, args) 
+
+
+@generated  function compute_τII_i(
```
albert-de-montserrat, line 205: I believe we don't need this function anymore

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+# Print info in the REPL
+include("CompositeRheologies_print.jl")
+
+
+
+function show(io::IO, g::AbstractComposite)
+    #println(io,"Composite rheology:   ")
+
+    # Compose a string with rheological elements, so we have an overview in the REPL
+    #str = create_rheology_string("",g)
```
albert-de-montserrat, line 105: remove line

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+    
+
+    return nothing
+end
+
+function show(io::IO, a::Parallel)
+    println(io,"Parallel:   ")  
+
+    # Compose a string with rheological elements, so we have an overview in the REPL
+    #str = create_rheology_string("",a)
```
albert-de-montserrat, line 117: remove line

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+
+@generated function getindex(p::CompositeRheology{T, N}, I::Int64) where {T,N}
+    quote
+        Base.@_inline_meta
+        Base.Cartesian.@nexprs $N i -> I == i && return p.elements[i]
+    end
+end
+
+# define rules to nondimensionalise this 
+function nondimensionalize(MatParam::Union{Parallel,CompositeRheology}, g::GeoUnits{TYPE}) where {TYPE}
```
albert-de-montserrat, line 72: Can this function and the one below be moved where all the other methods for `nondimensionalize()` and `dimensionalize()` are?

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+        par[id_parallel] .= 1
+    end
+    is_parallel =    SVector{n,Bool}(par)
+
+    return CompositeRheology{typeof(v), n, Npar, is_parallel, τ_it, P_it, λ_it}(v)
+end
+CompositeRheology(a,b...) = CompositeRheology( (a,b...,)) 
+CompositeRheology(a::Parallel) = CompositeRheology( (a,)) 
+
+@generated function getindex(p::CompositeRheology{T, N}, I::Int64) where {T,N}
```
albert-de-montserrat, line 64: I believe this is not needed anymore

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+end
+
+# Here we do need to do iterations
+function compute_εII_AD(v::Parallel, τII, args; tol=1e-6, verbose=false)
+    return local_iterations_τII_AD(v, τII, args; tol=tol, verbose=verbose)
+end
+
+
+# COMPUTE DEVIATORIC STRESS
+function compute_τII(v::CompositeRheology{T,N,0}, εII, args; tol=1e-6, verbose=false) where {T,N}
```
albert-de-montserrat, line 172: Is the type signature of `CompositeRheology` correct? It is defined as `CompositeRheology{T, N, Npar, is_parallel, τ_it, P_it, λ_it}`

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+    # A composite rheology case with no parallel element; iterations for τII
+    τII = local_iterations_εII(v, εII, args; tol=tol, verbose=verbose)
+    return τII
+end
+
+"""
+    τII = compute_τII(v::CompositeRheology{T,N}, εII, args; tol=1e-6, verbose=false)
+    
+"""
+function compute_τII(v::CompositeRheology{T,N}, εII, args; tol=1e-6, verbose=false, τ_initial=nothing, ε_init=nothing) where {T,N}
```
albert-de-montserrat, line 182: There's no need to add {T,N}  here

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+
+# VISCOSITY COMPUTATIONS
+
+""" 
+    η = computeViscosity_εII(v::Union{Parallel{T,N}, CompositeRheology{T,N}, AbstractConstitutiveLaw}, εII::_T, args; tol=1e-6, verbose=false)
+
+This computes the effective viscosity for a given input rheology `v` and strainrate `εII`
+"""
+function computeViscosity_εII(
+    v::Union{Parallel{T,N}, CompositeRheology{T,N}},
```
albert-de-montserrat, line 243: No need for {T,N}

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+    return η
+end
+
+""" 
+    η = computeViscosity_εII_AD(v::Union{Parallel{T,N}, CompositeRheology{T,N}, AbstractConstitutiveLaw}, εII::_T, args; tol=1e-6, verbose=false)
+
+This computes the effective viscosity for a given input rheology `v` and strainrate `εII`, while using AD if necessary
+"""
+function computeViscosity_εII_AD(
+    v::Union{Parallel{T,N}, CompositeRheology{T,N}, AbstractConstitutiveLaw},
```
albert-de-montserrat, line 265: No need for `{T,N}`

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+end
+
+# NONLINEAR ITERATION SCHEMES
+"""
+    τII =local_iterations_εII_AD(v::CompositeRheology{N,T,0}, εII::_T, args; tol=1e-6, verbose=false)
+
+Performs local iterations versus stress for a given total strain rate for a given `CompositeRheology` element that does NOT include `Parallel` elements
+"""
+@inline function local_iterations_εII(
+    v::CompositeRheology{N,T,0},
```
albert-de-montserrat, line 286: Again, is the type signature correct?

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+    return τII
+end
+
+"""
+    τII = local_iterations_εII_AD(v::CompositeRheology{N,T}, εII::_T, args; tol=1e-6, verbose=false)
+
+Performs local iterations versus stress for a given strain rate using AD
+"""
+@inline function local_iterations_εII_AD(
+    v::CompositeRheology{N,T},
```
albert-de-montserrat, line 331: Type signature?

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+
+
+
+"""
+    local_iterations_εII(c::CompositeRheology{T,N}, εII_total, args)
+
+This performs nonlinear Newton iterations for `τII` with given `εII_total` for cases where we have both serial and parallel elements.
+"""
+@inline function local_iterations_εII(
+    c::CompositeRheology{T,N,Npar, is_par},
```
albert-de-montserrat, line 456: Type signature

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+    end
+end
+
+
+"""
+    compute_εII_harmonic(v::Parallel{T,N}, TauII::_T, args)
+
+Computes the harmonic average of strainrate for a parallel element
+"""
+@generated function compute_εII_harmonic(
```
albert-de-montserrat, line 694: The function below is identical. Can't they be fused with a `Union`?

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+) where {T,N, _T}
+    quote
+        out = zero($_T)
+        Base.Cartesian.@nexprs $N i ->
+            out += inv(compute_εII(v.elements[i], TauII, args))
+        return inv(out)
+    end
+end
+
+@generated  function compute_εII_harmonic_i(
```
albert-de-montserrat, line 720: Function not needed anymore I believe

src/CompositeRheologies.jl
```diff
@@ -0,0 +1,794 @@
...
+    verbose=false
+) where {T,N, _T}
+    quote
+        out = zero(_T)
+        Base.Cartesian.@nexprs $N i ->
+            out += compute_εII_nonparallel(v.elements[i], TauII, args)
+    end
+end
+
+compute_εII_nonparallel(v::Any, TauII::_T, args) where {_T} = compute_εII(v, TauII, args)
```
albert-de-montserrat, line 751: No need for `Any`

src/ConstitutiveRelationships.jl
```diff
@@ -17,30 +17,48 @@ using StaticArrays
 
 const AxialCompression, SimpleShear, Invariant = 1, 2, 3
 
-abstract type AbstractConstitutiveLaw{T} <: AbstractMaterialParam end
+#abstract type AbstractConstitutiveLaw{T} <: AbstractMaterialParam end
+#abstract type AbstractComposite <: AbstractMaterialParam end
+
+precision(v::AbstractConstitutiveLaw) = (typeof(v).super).parameters[1]
```
albert-de-montserrat, line 23: Can be simplified as `precision(v::AbstractConstitutiveLaw{T})  where T= T`

albert-de-montserrat (changes requested):
Great job, this is an amazing PR !  Just some minor corrections are needed from my side

src/CompositeRheologies.jl
boriskaus: no it is still used

src/CompositeRheologies.jl
boriskaus: I tried doing that, but it complained about not knowing the `CompositeRheology` and `Parallel` structs. If you know how to fix that, it would be great!

src/CompositeRheologies.jl
boriskaus: I think this is handy to have, if working with `Parallel` and `CopositeRheology` structs so I would prefer to leave it in`

src/CompositeRheologies.jl
boriskaus: yes, but in this case we dont need the rest of the type signature. Is that a problem to not indicate it here?

src/CompositeRheologies.jl
boriskaus: correct, removed

src/CompositeRheologies.jl
boriskaus: as above

src/CompositeRheologies.jl
boriskaus: yes, done

src/ConstitutiveRelationships.jl
boriskaus: yes thanks

src/CompositeRheologies.jl
boriskaus: actually, it was indeed incorrect (should be {T,N,0} or the full one)

src/CompositeRheologies.jl
albert-de-montserrat: No, you are right. It's best to leave it like this for now given how everything is organised.

src/CompositeRheologies.jl
albert-de-montserrat: No that's fine, seems to work! I didn't know you could ommit the last 'n'-types in a signature. Good to know!

PR #41, opened 2022-09-30
additional testing & bugfix of CreepLaws

boriskaus:
This adds some basic testing for all entries in the databases and corrects a mistake in the DislocationCreep formulation


albert-de-montserrat (approved):
Looks great!

PR #46, opened 2022-10-04
Nils edit creep laws

Dr3ven:
I added all creep laws from the Bürgmann & Dresen (2008) paper supplementary, both diffusion and dislocation. All values have been checked with the original publications.


local_iters_anton.jl
```diff
@@ -0,0 +1,50 @@
+function Local_Iterations()
```
albert-de-montserrat, line 1: This whole file has been deprecated and doesn't exist anymore. You should remove it

src/CreepLaw/DiffusionCreep.jl
```diff
@@ -179,10 +183,10 @@ end
 @inline function compute_εII(
     a::DiffusionCreep, TauII::Quantity; T=1K, P=0Pa, f=1NoUnits, d=1e-3m, args...
 )
-    @unpack_units r, p, A, E, V, R = a
+    @unpack_units n, r, p, A, E, V, R = a
     FT, FE = a.FT, a.FE
 
-    ε = A * (TauII * FT) * f^(r) * d^(p) * exp(-(E + P * V) / (R * T)) / FE
+    ε = A * (TauII * FT)^(n) * f^(r) * d^(p) * exp(-(E + P * V) / (R * T)) / FE
```
albert-de-montserrat, line 189: you can use 'fastpow()' here instead of `^` for the powers. same in all the other functions

src/CreepLaw/DiffusionCreep.jl
```diff
@@ -255,36 +259,35 @@ Returns diffusion creep stress as a function of 2nd invariant of the strain rate
...
+    @unpack_val n, r, p, A, E, V, R = a
     FT, FE = a.FT, a.FE
 
     τ =
-        fastpow(A, -1 ) *
-                EpsII * FE *
-        fastpow(f, -r ) *
-        fastpow(d, -p ) *
-        exp((E + P * V) / (R * T)) / FT
+        fastpow(A, -1 / n) *
```
albert-de-montserrat, line 266: Division is 5-6 times slower than multiplication. You could precompute the inverse of `n` and multiply instead, i.e.
`_n = inv(v)` and then do for example `fastpow(A, -p*_n)` or whatever.

albert-de-montserrat (changes requested):
Looks good, some minor comments below

local_iters_anton.jl
Dr3ven: Done.

src/CreepLaw/DiffusionCreep.jl
Dr3ven: Done.

src/CreepLaw/DiffusionCreep.jl
Dr3ven: Done.

albert-de-montserrat (approved):

PR #47, opened 2022-10-04
Compressible rheologies

nicoberlie:
This PR adds a volumetric part to the rheology.

Routines are added to the Elasticity.jl file to compute the elastic volumetric strain rate and the associated pressure:

```julia
@inline function compute_εvol(
    a::ConstantElasticity, p::_T; p_old=zero(precision(a)), dt=one(precision(a)), kwargs...
) where {_T}
    @unpack_val Kb = a
    εvol_el = - (p - p_old) / (Kb * dt)
    return εvol_el
end

@inline function compute_p(
    a::ConstantElasticity, εvol::_T; p_old=zero(precision(a)), dt=one(precision(a)), kwargs...
) where {_T}
    @unpack_val Kb = a
    p = - Kb * dt * εvol + p_old

    return p
end
```

as well as the associated derivatives (```dεvol_dp``` and ```dp_dεvol```), the array formulation and the testing framework (I also added tests for the derivative routines already implemented).

I also added a function ```isvolumetric``` to recognize if a set rheology has a volumetric component. It should come in handy once we start using this with composite rheologies.


src/Elasticity/Elasticity.jl
```diff
@@ -72,6 +79,15 @@ function param_info(s::ConstantElasticity) # info about the struct
     return MaterialParamsInfo(; Equation=L"Constant elasticity")
 end
 
+function isvolumetric(a::ConstantElasticity)
+    @unpack_val ν = a
+    if ν == 0.5
```
albert-de-montserrat, line 84: just for the sake of having less lines of code, this can be just `return ν == 0.5 ? false : true`

src/Plasticity/Plasticity.jl
```diff
@@ -50,6 +51,15 @@ where ``\\dot{\\lambda}`` is a (scalar) that is nonzero and chosen such that the
 end
 DruckerPrager(args...) = DruckerPrager(convert.(GeoUnit, args)...)
 
+function isvolumetric(s::DruckerPrager)
+    @unpack_val Ψ = s
+    if Ψ == 0
```
albert-de-montserrat, line 56: `return Ψ == 0 ? false : true `

albert-de-montserrat (changes requested):

albert-de-montserrat (approved):

PR #49, opened 2022-10-06
Lf arheniusvisc

LukasFuchs:
Added rheology description for an arrhenius type viscosity in the CreepLaw routine. 
So far, this is only available for non-dimensional parameters. Scaling needs to be added for dimensional parameters


albert-de-montserrat (approved):
Thanks looks good!

PR #51, opened 2022-10-06
PlotDeformationMap routine

Dr3ven:
Created a `PlotDeformationMap()` routine which plots the deformation map (strainrate or stress vs. T/stress or T/strainrate) for a (composite) rheology. 
It is quite customisable and you can plot either stress, strainrate or effective viscosity:

```Julia
julia> v1 = SetDiffusionCreep("Dry Anorthite | Rybacki et al. (2006)")
julia> v2 = SetDislocationCreep("Dry Anorthite | Rybacki et al. (2006)")
julia> v=CompositeRheology(v1,v2)
julia> PlotDeformationMap(v, levels=100, colormap=:roma)
```
Screenshot 2022-10-25 at 17 29 44

```Julia
julia> PlotDeformationMap(v,  strainrate=false, viscosity=true, levels=Vector(18:.25:24), rotate_axes=true)
```
Screenshot 2022-10-25 at 17 30 59


boriskaus:
This looks very nice, but why don't you ask users to simply pass in `args`, instead of passing specific arguments such as grainsize and fugacity? 
That would make this work for future (combined) rheologies as well

src/Plotting.jl
```diff
@@ -903,4 +903,126 @@ function PlotStressTime_0D(
     end
 
     return fig, ax, Tau_II_MPa, t_vec
+
```
albert-de-montserrat, line 906: watch out, compilation is broken because you removed and `end` here to close the function above

src/Plotting.jl
```diff
@@ -903,4 +903,126 @@ function PlotStressTime_0D(
...
+                                              d::Quantity=1mm,
+                                              f::Quantity=1MPa,
+                                              P::Quantity=0MPa,
+                                              fig=nothing,
+                                              filename=nothing,
+                                              res=(1200, 900),)
+
+Creates a Deformation mechanism map (temperature vs. stress) for the given creep law with strain rate contours.
+"""
+function PlotDeformationMap(
```
albert-de-montserrat, line 922: You need to export this function at the top of the file

src/Plotting.jl
```diff
@@ -903,4 +903,126 @@ function PlotStressTime_0D(
...
+        ylabel="log10(σ) [MPa]",
+        )
+    c1 = contourf!(
+        ax,
+        T_plot, log_σ,
+        log_strainrates_D,
+        levels=11
+        )
+    contour!(
+        T_plot,
```
albert-de-montserrat, line 1009: This works fine, but it's better if you add the axis (i.e. `ax`) before `T_plot`, like you do just above for `c1`

src/Plotting.jl
```diff
@@ -903,4 +903,126 @@ function PlotStressTime_0D(
...
+            end
+        end
+    end
+
+    # Adjust arrays for plotting (get rid of units)
+    log_σ = [log10(σ[i].val) for i = 1:length(σ)]
+    T_plot = [T[i].val for i = 1:length(T)]
+
+    # Plotting with Makie
+    if isnothing(fig)
```
albert-de-montserrat, line 992: What is the intention of this check? If a `fig` is provided, you are plotting on top of the old plot, it may result in artifacts if something does not match the exact same position. I would remove `fig` from the `kwargs` list and make a new figure every time.

albert-de-montserrat (changes requested):
Thanks! The plot looks great. Some general comments:

1.  How do you actually produced that figure? I tried doing
```julia
v = SetMaterialParams(;
    Name="Mantle",
    Phase=1,
    CreepLaws=SetDiffusionCreep("Dry Anorthite | Rybacki et al. (2006)"),
)
 PlotDeformationMap((v,))
``` 
but it throws an error. Would be great if you add an example in the documentation!
2. It works for tuples only, but you are only plotting the deformation map of `Phase[1]` if I'm not mistaken. If this is the case, `Phase` should not be a tuple and you can do `Phase.CreepLaws` instead.
3.  You could add the colormap as a `kwarg` to give some extra flexibility to the user.
4. Is it possible to match the lines of `contour `with the with the contacts between the different colors  of `contourf`?

boriskaus:
I have made a few changes compared to the initial routine, which should make this more future-proof (will also work with new rheologies and not just for diffusion/dislocation creep).
Along the way, most changes requested by Albert were implemented. Instead of using `contourf` to add colors in the background, we now use `heatmap`, as the colors are otherwise cutoff at the lower/upper boundary.

boriskaus:
We do have a remaining issue, which is that plotting routines are only loaded once `GLMakie` is loaded. Similarly, documentation for those routines is only generated once this is the case. Yet, loading `GLMakie` does not work with `CI`, as the machines there do not have screens.

Not sure what the best way is to solve this.

albert-de-montserrat (approved):

PR #56, opened 2022-10-11
Remove println() calls when CUDA is active

albert-de-montserrat:
`println()` does not work in CUDA kerneles. This PR conditionally loads CUDA with `Requires` and switches off all the calls to `println` when CUDA has been loaded.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.3.18"
 
 [deps]
 BibTeX = "7b0aa2c9-049f-5cec-894a-2b6b781bb25e"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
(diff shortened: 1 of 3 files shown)


boriskaus (approved):
great!

luraess:
If needed there is also `@cuprintln` for printing info from within CUDA kernels.

albert-de-montserrat:
@luraess Indeed, but I had problems trying to overload `println` with `@cuprintln`. I guess I can make it work messing around with `eval` calls, but I don't think we want to e.g. print the iterations of every Newton solve in every GPU thread anyway.

luraess:
Sure - was just my 5 cents ;-) I guess it's optimal as you did!

PR #57, opened 2022-10-12
Adm-improve-#56

albert-de-montserrat:
This PR improves [#56](https://github.com/JuliaGeodynamics/GeoParams.jl/pull/56). I realised it made no sense to conditionally load CUDA and add it as a dependency.


boriskaus:
will this cause issues if you run on AMD machines?

albert-de-montserrat:
You are right didn't think about it. I added a `isGPU()` check instead, that disables printing if either CUDA or AMDGPU is loaded.

boriskaus (approved):
nice, thanks!

PR #58, opened 2022-10-12
1D Strength Envelope

ArneSpang:
Added a function that creates a GUI where you can plot a 1D strength envelope.
Inputs: A tuple of different materials (including phase, density, creeplaw and plasticity)
            A Vector with the Thicknesses of the layers
            A temperature profile type (constant, linear=default, halfspace-cooling)

Output: A figure showing the strength envelope with sliders to adjust the temperature and background strain rate

Also added documentation and tests


boriskaus:
this is nice very nice. Yet, it would be fantastic to not just restrict this to linear temperature structures, but instead also allow half-space cooling thermal structures (with given age; perhaps as slider?) Could that be added?

src/Plotting/StrengthEnvelope.jl
```diff
@@ -0,0 +1,319 @@
...
+                   SetMaterialParams(Name="LC", Phase=3, Density=PT_Density(ρ0=2900kg/m^3, α=3e-5/K, β=1e-10/Pa), CreepLaws = SetDislocationCreep("Maryland strong diabase | Mackwell et al. (1998)"), Plasticity = DruckerPrager(ϕ=30.0, C=10MPa)));
+julia> Thickness = [15,10,15]*km;
+
+julia> StrengthEnvelope(MatParam, Thickness, LinearTemp())
+```
+"""
+function StrengthEnvelope(MatParam::NTuple{N, AbstractMaterialParamsStruct}, Thickness::Vector{U}, TempType::AbstractThermalStructure=LinearTemp(), mode::String="normal") where {N, U}
+
+    # hardcoded input
+    nz        = 101
```
boriskaus, line 186: can you make this a changeable part of the input function?
Do that with:
```Julia
function StrengthEnvelope(MatParam::NTuple{N, AbstractMaterialParamsStruct}, Thickness::Vector{U}, TempType::AbstractThermalStructure=LinearTemp(), mode::String="normal"; nz=101)
```
Default will be 101 in that case, but you have the possibility to increase it

boriskaus (changes requested):

ArneSpang:
Ok, so all tests fail because it does not find Makie. So does Makie need to be a mandatory dependency after all?

src/Plotting/StrengthEnvelope.jl
boriskaus: > > this is nice very nice. Yet, it would be fantastic to not just restrict this to linear temperature structures, but instead also allow half-space cooling thermal structures (with given age; perhaps as slider?) Could that be added?
> 
> That is already part of it

Sorry, I had missed that.

boriskaus:
> Ok, so all tests fail because it does not find Makie. So does Makie need to be a mandatory dependency after all?

Now we don't want that, as it is a very big dependency and not required for computations. What you can do is add this as a dependency only for the testing (I forgot the details of how to do this, but it is explained on the package manager part of the Julia webpage, as far as I recall).

ArneSpang:
> > Ok, so all tests fail because it does not find Makie. So does Makie need to be a mandatory dependency after all?
> 
> Now we don't want that, as it is a very big dependency and not required for computations. What you can do is add this as a dependency only for the testing (I forgot the details of how to do this, but it is explained on the package manager part of the Julia webpage, as far as I recall).

There is a Project.toml in the test directory aswell. I can try putting Makie there when I push the nz change

boriskaus:
indeed try that. Please remember to also add versions for all the packages you add; that is required for the Julia registry.

ArneSpang:
> indeed try that. Please remember to also add versions for all the packages you add; that is required for the Julia registry.

Seems like that worked but now there are other issues with Makie:

Warning:     OpenGL/GLFW wasn't loaded correctly or couldn't be initialized.
│     This likely means, you're on a headless server without having OpenGL support setup correctly.

ERROR: LoadError: InitError: Exception[GLFW.GLFWError(GLFW.PLATFORM_ERROR, "X11: The DISPLAY environment variable is missing"), ErrorException("glfwInit failed")]

No idea what can be done about that. Maybe packages that can open windows just don't work in this testing framework?

boriskaus:
perhaps you can use a setting which does allow doing this? Or use CairoMakie instead of GLMakie for these plots? I assume that the Makie help will clarify this

src/Plotting/StrengthEnvelope.jl
```diff
@@ -0,0 +1,319 @@
...
+Sets a constant temperature inside the box
+Parameters
+===
+- T : the value
+"""
+@with_kw_noshow mutable struct ConstantTemp <: AbstractThermalStructure
+    T = 1000
+end
+
+function GP_Compute_ThermalStructure(Z, s::ConstantTemp)
```
albert-de-montserrat, line 21: This can simply be a one-liner `GP_Compute_ThermalStructure(Z, s::ConstantTemp) = fill(s.T, length(Z)`

src/Plotting/StrengthEnvelope.jl
```diff
@@ -0,0 +1,319 @@
...
+
+"""
+    ConstantTemp(T=1000)
+    
+Sets a constant temperature inside the box
+Parameters
+===
+- T : the value
+"""
+@with_kw_noshow mutable struct ConstantTemp <: AbstractThermalStructure
```
albert-de-montserrat, line 17: Does this really need to be mutable? If this number really changes it's probably better to make a normal immutable struct and just reallocate it doing `ConstantTemp(T)`

src/Plotting/StrengthEnvelope.jl
```diff
@@ -0,0 +1,319 @@
...
+"""
+    LinearTemp(Ttop=0, Tbot=1000)
+    
+Set a linear temperature structure from top to bottom
+Parameters
+===
+- Ttop : the value @ the top
+- Tbot : the value @ the bottom
+"""
+@with_kw_noshow mutable struct LinearTemp <: AbstractThermalStructure
```
albert-de-montserrat, line 37: Same as before, these numbers probably do not mutate.

src/Plotting/StrengthEnvelope.jl
```diff
@@ -0,0 +1,319 @@
...
+
+function GP_Compute_ThermalStructure(Z, s::LinearTemp)
+    @unpack Ttop, Tbot  = s
+
+    Temp  = zeros(Float64, length(Z))
+
+    dz    = Z[end]-Z[1];
+    dT    = Tbot - Ttop
+
+    Temp .= abs.(Z./dz).*dT .+ Ttop
```
albert-de-montserrat, line 50: You don't really need to initialize `Temp` above,  just do `Temp = @.  abs(Z/dz)*dT + Ttop`

luraess:
CairoMakie works pretty robust (also on servers etc...), but is not interactive.

src/Plotting/StrengthEnvelope.jl
```diff
@@ -0,0 +1,319 @@
...
+        Temp = T[i]
+        args = (T=Temp, P=Pres)
+        Mat  = MatParam[Phases[i]]
+        τ[i] = compute_τII(Mat.CreepLaws[1], ε, args)
+
+        F    = compute_yieldfunction(Mat.Plasticity[1], P=Pres, τII=τ[i])
+        if F > 0
+            ϕ = Mat.Plasticity[1].ϕ.val
+            c = Mat.Plasticity[1].C.val
+            τ[i] = Pres * sind(ϕ) + cosd(ϕ) * c
```
albert-de-montserrat, line 146: `sind(ϕ)` and `cosd(ϕ)` are already inside the Drucker-Prager struct. They are called `Mat.Plasticity[1].sinϕ` and  `Mat.Plasticity[1].cosϕ`

src/Plotting/StrengthEnvelope.jl
```diff
@@ -0,0 +1,319 @@
...
+    z         = collect(0:dz:Inter[end])
+
+    # distribute phases
+    Phases    = ones(Int64, nz) * MatParam[1].Phase
+    for i = 1 : nLayer - 1
+        Phases[z .> Inter[i]] .= MatParam[i+1].Phase
+    end
+
+    # check temperature structure
+    if typeof(TempType) == LinearTemp
```
albert-de-montserrat, line 208: Since you are checking for the type of `TempType`, you could remove all these branching by using multiple dispatch

albert-de-montserrat (commented):
Very nice @ArneSpang  thanks!

src/Plotting/StrengthEnvelope.jl
ArneSpang: > `sind(ϕ)` and `cosd(ϕ)` are already inside the Drucker-Prager struct. They are called `Mat.Plasticity[1].sinϕ` and `Mat.Plasticity[1].cosϕ`

Thanks for all the suggestions. Implemented all of the above and checked functionality.

src/Plotting/StrengthEnvelope.jl
ArneSpang: > Since you are checking for the type of `TempType`, you could remove all these branching by using multiple dispatch

I could but maybe the compute temperature functions will at some point be used by some other functions as well and I would like to keep them as general as possible. I agree that the current way is a bit long but I like that it is very easy to follow.

src/Plotting/StrengthEnvelope.jl
ArneSpang: > GLMakie works on GPUs. As @boriskaus points out, try CairoMakie as an alternative, but double-check first if all the interactive GUI widgets also work (or at least don't crash)

I checked and as Ludo pointed out, CairoMakie cannot create the interactive plot. As the test does not need to be interactive, I thought I could put this at the start of the StrengthEnvelope function:

if mode == "normal"
        using GLMakie
elseif mode == "test"
        using CairoMakie
end

but this also crashes because the Makie functionalities are not loaded yet when it tries to compile the function.

ArneSpang: > I am wondering whether we should or can even have a test for this. Looking at how Makie does testing it wasn't immediately obvious to me whether they are even doing this for widgets...

I was also wondering if I should just remove the test and hope that it never breaks or instead write a function that is a clone but does not use any of the makie functionality and only default values for testing. But this would be annoying to maintain.

src/Plotting/StrengthEnvelope.jl
boriskaus: you could make a separate computational routine that does all the work required here, and simply call this routine from the plotting routine. In that case you don't duplicate code and it is easier to maintain (plus can be tested)

ArneSpang: > you could make a separate computational routine that does all the work required here, and simply call this routine from the plotting routine. In that case you don't duplicate code and it is easier to maintain (plus can be tested)

So I ended up going down this road and stumbled upon a general issue which causes problems for the halfspace-cooling case:
Why does this work? 
julia> 500m + 1km
1500 m

and this does not?
julia> 500m / 1km
500.0 m km^-1.0

boriskaus: Good point. These type of calculations use the [Unitful.jl](https://github.com/PainterQubits/Unitful.jl) package, and it seems that this package does not support that:

```julia
julia> using Unitful
julia> 1000u"km"
1000 km
julia> 1u"km" + 5u"m"
1005 m
julia> 1u"km" / 5u"m"
0.2 km m⁻¹
julia> 1u"km" * 5u"m"
5 m km
julia> 1u"km" - 5u"m"
995 m
```

ArneSpang:
So I split the function in two. StrengthEnvelopeComp is the computational part that is independent of Makie, loaded every time and has tests. StrengthEnvelopePlot is the interactive plotting routine that is only loaded when the user loads Makie and has no tests.
I ended up putting two lines of uconvert into the Halfspacecooling function to make it more stable when used with units.

So I think it is ready to be pulled. But generally I think it is silly that we cannot divide meter by kilometer and similar things. Do you know anyone that works on Unitful or should I contact them to ask if this is going to be added at some point?

ArneSpang:
I did look into Unitful again and found a function NoUnits() which does take care of obvious unit cancelling and gives an error when units actually don't cancel, so I use that now.

julia> 500m/1km
500.0 m km^-1.0

julia> NoUnits(500m/1km)
0.5

julia> NoUnits(500/1km)
ERROR: DimensionError:  and km^-1.0 are not dimensionally compatible.

boriskaus (approved):

PR #60, opened 2022-10-19
Another attempt to improve documentation

ArneSpang:
Realized that documentation that is in the plotting routines will never be build so I moved it to a different file so it actually shows up. Hopefully this is the last one. The documentation does unfortunately not build on the fork. So I have to keep trying with pull requests.


albert-de-montserrat (approved):
Great thanks! Is it ready to merge then?

ArneSpang:
> Great thanks! Is it ready to merge then?

yes

PR #61, opened 2022-10-20
Addition of volumetric rheologies to composite rheologies

nicoberlie:
Addition to volumetric indications to the name of the rheology as well as functions to compute pressure and dependencies for composite rheologies:

```Julia
function compute_p_τII()
function local_iterations_εvol()
function compute_p_harmonic()
```

Tests and parallel case not finished and the compressible plasticity is a work in progress but Boris wants to contribute, hence the draft.


src/CompositeRheologies.jl
```diff
@@ -132,6 +138,10 @@ function show(io::IO, a::Parallel)
     return nothing
 end
 
+function isvolumetric(c::NTuple{N,Any}) where {N}
```
albert-de-montserrat, line 141: might be slightly better to define this function as `function isvolumetric(c::NTuple{N,T}) where {N,T<:AbstractCreepLaw}` - or whatever the appropriate type is.

src/CompositeRheologies.jl
```diff
@@ -206,13 +225,136 @@ function compute_εII_AD(v::Parallel, τII, args; tol=1e-6, verbose=false)
...
+
+@generated  function compute_εvol(
+    v::CompositeRheology{T,N}, 
+    p::Quantity, 
+    args; 
+    tol=1e-6, verbose=false
+) where {T,_T,N}
+    quote
+        Base.@_inline_meta
+        εvol = 0/s
```
albert-de-montserrat, line 261: `zero(T)/s`

src/CompositeRheologies.jl
```diff
@@ -206,13 +225,136 @@ function compute_εII_AD(v::Parallel, τII, args; tol=1e-6, verbose=false)
...
+
+
+"""
+    p,τII = compute_p_τII(v::CompositeRheology, εII, εvol, args; tol=1e-6, verbose=false) 
+
+This updates pressure `p` and deviatoric stress invariant `τII` in case the composite rheology structure has no volumetric elemnts, but may contain plastic or parallel elements. 
+In that case, pressure is not updated (`args.P` is used instead).     
+"""
+function compute_p_τII(
+    v::CompositeRheology{T,N,
```
albert-de-montserrat, line 311: As a matter of format and style, align `v` with the arguments below and keep its type signature in a single line.

src/CompositeRheologies.jl
```diff
@@ -206,13 +225,136 @@ function compute_εII_AD(v::Parallel, τII, args; tol=1e-6, verbose=false)
...
+end
+
+"""
+    p,τII = compute_p_τII(v::CompositeRheology, εII, εvol, args; tol=1e-6, verbose=false) 
+
+This updates pressure `p` and deviatoric stress invariant `τII` in case the composite rheology structure has volumetric components and has volumetric plasticity
+The 'old' pressure should be stored in `args` as `args.P_old`   
+"""
+function compute_p_τII(
+    v::CompositeRheology{T,N,
```
albert-de-montserrat, line 336: fix format

src/CompositeRheologies.jl
```diff
@@ -794,39 +1005,235 @@ end
...
+        if is_plastic[i] && !is_par[i]
+            # parallel plastic element
+            j += 1
+            x[j] = 0    # λ̇  
+            j += 1
+            x[j] = p_initial    # initial guess for pressure
+        end
+            
+    end
+#=
```
albert-de-montserrat, line 1159: remove all this commented block

src/CompositeRheologies.jl
```diff
@@ -794,39 +1005,235 @@ end
...
+    end    
+    verbose && println("τII guess = $τ_initial \n  P guess = $p_initial")
+    
+    x    = @MVector zeros(_T, n)
+    x[1] = τ_initial
+
+    j = 1;
+    for i=1:N
+
+        if is_plastic[i] && is_par[i]
```
albert-de-montserrat, line 1143: Could you try to use the Static.jl package to replace this branching by multiple dispatch (like [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/c93f5f8d2d3152ad96790258e540f725cdf25753/src/CompositeRheologies.jl#L749)) and see if there's a performance gain?

src/CompositeRheologies.jl
```diff
@@ -794,39 +1005,235 @@ end
...
+        r[1]   = εII_total - compute_εII_elements(c,τ,args)     
+        J[1,1] = dεII_dτII_elements(c,τ,args);               
+        
+        r[n]   = εvol_total - compute_εvol_elements(c,P,args)     
+        J[n,n] = dεvol_dp(c,P,args);               
+        
+        # Add contributions from plastic elements
+        fill_J_plastic!(J, r, x, c, args)
+
+        #@show x r J
```
albert-de-montserrat, line 1205: remove line

src/CompositeRheologies.jl
```diff
@@ -794,39 +1005,235 @@ end
...
+end
+
 @generated function ∂Q∂τII(
-    v::Parallel{T, N,  Nplast, is_plastic}, τ::_T
+    v::Parallel{T, N,  Nplast, is_plastic}, τ::_T, args
 ) where {_T,T, N,  Nplast, is_plastic}
     quote
         Base.@_inline_meta
-        Base.Cartesian.@nexprs $N i -> is_plastic[i] == true && return ∂Q∂τII(v[i],τ)
+        Base.Cartesian.@nexprs $N i -> is_plastic[i] == true && return ∂Q∂τII(v[i],τ, args)
```
albert-de-montserrat, line 1227: Add `return zero(_T)` at the end

src/CompositeRheologies.jl
```diff
@@ -794,39 +1005,235 @@ end
...
+    end
+end
+
+@generated function ∂Q∂P(
+    v::Parallel{T, N,  Nplast, is_plastic}, P::_T, args
+) where {_T,T, N,  Nplast, is_plastic}
+    quote
+        Base.@_inline_meta
+        Base.Cartesian.@nexprs $N i -> is_plastic[i] == true && return ∂Q∂P(v[i],P, args)
```
albert-de-montserrat, line 1236: Add `return zero(_T)` at the end

src/CompositeRheologies.jl
```diff
@@ -206,13 +225,136 @@ function compute_εII_AD(v::Parallel, τII, args; tol=1e-6, verbose=false)
...
+    ) where {T, N, _T, Npar, is_parallel, Nplastic, is_plastic, Nvol, is_vol}
+
+    # A composite rheology case that may have volumetric elements, but the are not 
+    # tightly coupled, so we do NOT perform coupled iterations.
+    out = local_iterations_εvol_εII(v, εII, εvol, args; tol=tol, verbose=verbose)
+
+    τII = out[1]
+    P = out[end]
+
+    return P,τII, out[2:N]
```
albert-de-montserrat, line 354: is this necessary? Couldn't this function just return `out`?

albert-de-montserrat (changes requested):

src/CompositeRheologies.jl
boriskaus: in fact, we won't even need this function as we can use broadcasting:
```Julia
isvolumetric.(xxx)
```

src/CompositeRheologies.jl
boriskaus: agree with then aligning, but keeping the type signature in a single line makes it less readable in my opinion.

src/CompositeRheologies.jl
boriskaus: we expect  P,τII (in that order) in the following routine; in `out` it is ordered differently. we can ofcourse rearrange it in a follow-up routine. changing the order within out would also be possible somehow, but we would end up with having to rewrite parts of the routine (which would make this not reusable in the non-dilatant plasticity routine)

albert-de-montserrat: As long as `out[2:N]` does not allocate, happy with it.

src/CompositeRheologies.jl
boriskaus: hmm, this doesn't actually seem to work (tests fail if I add this)

src/CompositeRheologies.jl
boriskaus: doesn't:
```Julia
julia> εxx,εzz  = 7e-15, -6.8e-15
julia> εII  = sqrt(0.5*(εxx^2 + εzz^2))
julia> εvol = εxx + εzz
julia> args = (T = 900.0, d = 0.0001, τII_old = 700000.0, dt = 8.0e9, P = 0.0, P_old = 1e6)
julia> c16= CompositeRheology(SetConstantElasticity(G=1e10, Kb=2e11), LinearViscous(η=1e20), DruckerPrager_regularised(C=3e5, Ψ=10, η_vp=1e19))
--/\/\/----⟦▪̲̅▫̲̅▫̲̅▫̲̅----▬▬▬__--
julia> @btime compute_p_τII($c16, $εII, $εvol, $args)
  52.581 ns (0 allocations: 0 bytes)
(1.2583092027089968e6, 909776.9097816481, (2.081468729181813e-15, 1.2583092027089968e6))
```

albert-de-montserrat (approved):
Great job!

PR #65, opened 2022-10-26
Remove redundant CPU checks

albert-de-montserrat:
Checking if CUDA is active is really slow and dominates the total time spend in the local iteration kernels.

image

 This PR removes all the redundant checks and considerably speeds up these kernels:

```julia-repl
# before changes
@btime computeViscosity_εII($v, $ε̇ii, $args) 
 310.163 ns (0 allocations: 0 bytes)

# after changes
@btime computeViscosity_εII($v, $ε̇ii, $args) 
  83.126 ns (0 allocations: 0 bytes)
```

### DIFF
--- a/src/CompositeRheologies/NonlinearIterations.jl
+++ b/src/CompositeRheologies/NonlinearIterations.jl
@@ -19,7 +19,8 @@ Performs local iterations versus stress for a given total strain rate for a give
     # Initial guess
     τII = compute_τII_harmonic(v, εII, args)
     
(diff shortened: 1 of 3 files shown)


boriskaus:
if it is this slow, why don't we expand the type signature of the `Parallel` and `CompositeRheology` structs to include an `isGPU` flag. In that case, the costly part would only be done while constructing the `structs`, not?

albert-de-montserrat:
Worth giving it a try. I will work on it

boriskaus (approved):
great!

PR #66, opened 2022-11-15
Fix type stability for SetMaterialParams

albert-de-montserrat:
Fix a type instability when instantiating material structs with `SetMaterialParams`.  Not critical for simulations, but this was a bit annoying, plus there's a significant improvement:
```julia
foo() = SetMaterialParams(;
    Name="Matrix",
    Phase=1,
    Density=ConstantDensity(),
)
```
```julia-repl
@btime foo() # pre-PR
     1.120 μs (14 allocations: 1.05 KiB) 

@btime foo() # this PR (updated with Static.jl)
      117.985 ns (3 allocations: 144 bytes) 
```


boriskaus (approved):
Great!

PR #69, opened 2022-11-29
Stress Jacobians

albert-de-montserrat:
Adds the `compute_dτijdεij` family of methods that return both the stress tensor and its jacobian


boriskaus:
This is a great PR as well! The idea is that the resulting Jacobians can directly be inserted in a (nonlinear) implicit finite element code?

PR #73, opened 2023-01-31
Parameter output tables for CompositeRheology

Dr3ven:
I adjusted the code to work for CompositeRheologys. Below an example table for LaTeX and Markdown with a CompositeRheology. Also added test_Table cases for CompositeRheology.
Phase:
```julia
v1 = SetDiffusionCreep("Dry Anorthite | Rybacki et al. (2006)")
c1 = CompositeRheology(v1, SetDislocationCreep("Diabase | Caristan (1982)"), LinearViscous(), v1)
MatParam = (SetMaterialParams(Name="Viscous Matrix", Phase=1, Density=ConstantDensity(),CreepLaws = SetDislocationCreep("Quartz Diorite | Hansen & Carter (1982)")),
            SetMaterialParams(Name="Viscous Sinker", Phase=2, Density= PT_Density(),CompositeRheology = c1),
            SetMaterialParams(Name="Viscous Bottom", Phase=3, Density= PT_Density(),CreepLaws = SetDiffusionCreep("Dry Anorthite | Rybacki et al. (2006)")))
```

LaTeX output:
output_table_example_latex
...plus an References.bib file for the citations.

Markdown output:
output_table_example_markdown


src/Tables.jl
```diff
@@ -1,92 +1,396 @@
...
+Phase2Dict() puts all parameters of a phase s in a dict.
+"""
 function Phase2Dict(s)
-    # Dict has Key with Fieldname and Value with Tuple(value, symbol, unit)
-    fds = Dict{String,Tuple{String,String,String,String}}()
+    # Dict has Key with Fieldname and Value with Tuple(value, symbol, flowlaw keyword, number of phases, (Disl+Diff) or Lin, Disl or Diff)
+    fds = Dict{String,Tuple{String,String,String,String,String,String}}()
     refs = Dict{String,Tuple{String,String,String}}()
-    # Descriptions for every parameter that could occur in the table and their corresponding variable name(s) that is used in GeoParams
+    if !(typeof(s) <: Tuple)
```
albert-de-montserrat, line 39: This could be a function with a couple of methods in `Utils.jl`, may become handy eventually. Something like:
```julia
make_tuple(x) = (x,)
make_tuple(x::Tuple) = x
```

src/Tables.jl
```diff
@@ -177,36 +486,105 @@ function Dict2LatexTable(d::Dict, refs::Dict)
...
     symbs = unique(sort(symbs))
 
     # Creates columnwise output for all parameters of the input phase
     for symbol in symbs
-        # Sets parametername and variable
-        Table *= " " * string(desc[symbol]) * " & " * "\$" * symbol * "\$"
+        # Sets parametername and variable (symbol)
+        # Order of signs in symbol: 1. "_", if already "_" in there -> 2. "^"
+        # If no underscore in symbol
+        if (maximum(endswith.(symbol,["0","1","2","3","4","5","6","7","8","9"])) && !occursin("_", symbol)) || (!occursin("\\", symbol) && length(symbol) > 1 && !occursin("_", symbol))
```
albert-de-montserrat, line 523:
```julia
x = ["0","1","2","3","4","5","6","7","8","9"]
y = [string(i) for i in 0:9]
x==y # true
```
and similar in line 551

src/Tables.jl
```diff
@@ -1,92 +1,396 @@
...
+        s = (s,)
+    end
     phasecount = length(s)
     k = 1
     flowlawcount = 0
     # Checks all Phases 
     for i in 1:phasecount
         fieldnames = propertynames(s[i])
-        # Goes through all fields of a phase that are not "Name"
+        global Diff = 0
```
albert-de-montserrat, line 48: Maybe I'm missing it, but do `Diff`, `Disl` and `Lin` need to be `global`? If there's not a specific reason, avoid using global variables

src/Tables.jl
```diff
@@ -256,24 +636,573 @@ function Dict2LatexTable(d::Dict, refs::Dict)
...
+    if !(typeof(s) <: Tuple)
+        s = (s,)
+    end
+    phasecount = length(s)
+    k = 1
+    flowlawcount = 0
+    # Checks all Phases 
+    for i in 1:phasecount
+        fieldnames = propertynames(s[i])
+        global Diff = 0
```
albert-de-montserrat, line 678: Again, there's no need for global variables

src/Tables.jl
```diff
@@ -256,24 +636,573 @@ function Dict2LatexTable(d::Dict, refs::Dict)
...
+        # Sets parametername and variable
+        if length(symbol) > 1
+            # Checks if Unicode chars in combination with a number etc. (e.g "α0") are used which apparently do not have a second index but only first and third
+            # and checks if symbol has another number in it than 0
+            if maximum(endswith.(symbol,["0","1","2","3","4","5","6","7","8","9"])) && !occursin("_", symbol)
+                if length(unidecode(symbol)) > 2 && occursin("0", symbol)
+                    Table *= " " * string(desc[symbol]) * " | " * symbol[1] * "~" * symbol[3] * "~"
+                elseif length(unidecode(symbol)) <= 2 && occursin("0", symbol)
+                    Table *= " " * string(desc[symbol]) * " | " * symbol[1] * "~" * symbol[2] * "~"
+                elseif length(unidecode(symbol)) > 2 && maximum(occursin.(["1","2","3","4","5","6","7","8","9"], symbol))
```
albert-de-montserrat, line 1100: `maximum(occursin(string(i), symbol) for i in 1:9)` is more idiomatic and does not allocate an intermediate array

src/Tables.jl
```diff
@@ -256,24 +636,573 @@ function Dict2LatexTable(d::Dict, refs::Dict)
...
+    end
+    symbs = unique(sort(symbs))
+
+    # Creates columnwise output for all parameters of the input phase
+    for symbol in symbs
+        # Sets parametername and variable
+        if length(symbol) > 1
+            # Checks if Unicode chars in combination with a number etc. (e.g "α0") are used which apparently do not have a second index but only first and third
+            # and checks if symbol has another number in it than 0
+            if maximum(endswith.(symbol,["0","1","2","3","4","5","6","7","8","9"])) && !occursin("_", symbol)
```
albert-de-montserrat, line 1095: `maximum(endswith(symbol, string(i),) for i in 1:9)` a better choice

albert-de-montserrat (commented):
Thanks a lot Nils! Looks really cool :)  

I couldn't find where you make the first column of the LaTeX table, but could you change the name of `Linear Viscous` to `Linear viscosity`?

src/Tables.jl
Dr3ven: I tried it without the global. It only worked when these variables were global. I also wondered why they need to be global but I couldn't figure it out.

boriskaus:
This is really cool! 
Could you perhaps a chapter in the help documents to explain how to do this? Also, the README would need an update, to mention it is there (now it states it is scheduled, I believe)

src/Tables.jl
Dr3ven: Making them local also doesn't work: "UndefVarError: Disl not defined".

Dr3ven:
Will do.

src/Tables.jl
Dr3ven: Not quite sure what the second variant of your given example should do, when x is already a tuple .
Also, isn't it easier to just use `(x,)` by its own since `make_tuple` would do the same but with more typing?

Dr3ven:
Doc and README should now be updated.

docs/src/man/tables.md
```diff
@@ -0,0 +1,36 @@
+# Tables
+
+There are two differnt formats of output tables that can be produced:
```
boriskaus, line 3: Typo

boriskaus (changes requested):
Only a small typo left

boriskaus (approved):

src/Tables.jl
albert-de-montserrat: If you define those 2 methods, you can basically remove the branching. If `x` is already a tuple, then `make_tuple` does nothing and returns `x` itself, whereas if it's not a tuple, it makes it one.  The whole point is that in Julia, you can avoid introducing branching that checks on types by making use of multiple dispatch. And in the end I would say doing `make_tuple(s)` is less verbose and more readable than checking for its type.

albert-de-montserrat (approved):

src/Tables.jl
Dr3ven: That would defeat the purpose of those variables since they count the occurences of DislCreep, DiffCreep and LinVisc per phase.
You marked the problem as resolved. Is that correct or just a misclick?

src/Tables.jl
Dr3ven: Ah, ok. I implemented it.

Dr3ven:
How can I use a function from Utils.jl without it being a module? Using only works when its a module if I'm not mistaken and include is too much for just using one function from Utils.jl.

albert-de-montserrat:
`Utils.jl` is included in GeoParams. You can import whatever function `foo` doing `using GeoParams: foo`

src/Tables.jl
albert-de-montserrat: I think for now we can leave it as it is since it's working.

PR #77, opened 2023-03-14
first commit to implement Melt dependent viscosity

Iddingsite:
Here a first try to implement Melt dependent viscosity based on Schmeling et al., 2012  (not working right now because it doesn't take into account args argument). Also needs tests.


albert-de-montserrat:
I guess this PR is dead and melt-dependant viscosity has been already implemented in #164. I will close this.

PR #87, opened 2023-05-03
Local density function for multi-phases rheology given volume fractions

albert-de-montserrat:
There was an in-place function `compute_density!(rho, Mat_tup1, PhaseRatio, args)` to do this but we are missing the one to do it locally, which is more convinient for our codes. To avoid name redunancy and ambiguity, I think it is better to use a different name for local volume fractions kernels, e.g. `compute_density_ratio(PhaseRatio, rheologies, args)` here.

### DIFF
--- a/src/Density/Density.jl
+++ b/src/Density/Density.jl
@@ -18,6 +18,7 @@ abstract type AbstractDensity{T} <: AbstractMaterialParam end
 
 export compute_density,        # calculation routines
     compute_density!,       # in place calculation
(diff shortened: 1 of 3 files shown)


boriskaus:
fair - does this allocate.

albert-de-montserrat:
No :) 
```julia-repl
julia> @btime  compute_density_ratio($PhaseRatio, $rheologies, $args)   
  5.200 ns (0 allocations: 0 bytes)
```

PR #96, opened 2023-08-25
Performance of viscosity kernels

albert-de-montserrat:
Improves performance of viscosity calculations by ~10%.


boriskaus:
We should probably add a test for 1.9 as well, not? And are the fails in nightly something to worry about? Would be bad if the code is broken when 1.10 is released.

albert-de-montserrat:
Should definetly test against 1.9.X. Testing nightly is a bit more annoying but I would not remove the test, I can take care of it later on.

boriskaus (approved):

boriskaus:
They released an alpha version of 1.10 now; I think there is a way to call that from the CI

PR #98, opened 2023-09-12
fix typos and add spell check

ranocha:
This setup will run a spell checker for every PR. I fixed the version of the spell checker to avoid letting PRs fail due to updates of it. Dependabot will automatically create PRs with updates of the spell checker. To reduce their frequency, I set up Dependabot to run monthly instead of weekly.

If you like this, we can set it up for other repos as well.


boriskaus (approved):
this is great, thanks!

Issue #104, opened 2023-09-13
Functionality using Plots.jl is broken


ranocha:
There is some code using Plots.jl explicitly, e.g.,

https://github.com/JuliaGeodynamics/GeoParams.jl/blob/234c1ad2dc502e504355d1312312d2f7903f56c3/src/Plotting/Plotting.jl#L748-L767

However, the optional plotting has been changed to GLMakie.jl. Thus, parts like these need to be fixed.

Issue #103, opened 2023-09-13
Add code coverage reports


ranocha:
It would be helpful to upload the code coverage reports from CI runs to coveralls/codecov to be able to see whether all parts of the code are covered by tests.

PR #105, opened 2023-09-13
Use a package extension for GLMakie-based functionality

ranocha:
I updated the optional plotting code from Requires.jl to package extensions for newer Julia versions. I kept backwards compatibility by still using Requires.jl for old Julia versions. This cuts the loading time `using GLMakie, GeoParams` for me by something between 0.5 and 1.5 seconds. However, there is still a lot of recompilation left, so this should be investigated in the future.

Caveats:
- I did not test this - except checking that `using GLMakie, GeoParams` works on Julia v1.8 and v1.9. I do not know whether the plotting routines are tested in CI (see #103) but I will assume that everything is fine if CI passes.
- There are still some remnants of a previos version using Plots.jl (see #104). I did not update them to GLMakie.jl.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -56,4 +56,3 @@ julia = "1.7"
 
 [extras]
 GLMakie = "e9467ef8-e4e7-5192-8a1a-b1aee30e663a"
-Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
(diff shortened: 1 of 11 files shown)


albert-de-montserrat (approved):
Thanks! Looks good to me

ranocha:
Ready to be merged?

boriskaus:
Thanks for the changes and apologies for the delay in merging

PR #99, opened 2023-09-12
2 different Peierls Creep rheologies and Grain Boundary Sliding (GBS)

Dr3ven:
I implemented 2 different Peierls creep laws. The NonLinearPeierlsCreep has a power law part in its equation, which makes it differ from the "normal" Peierls creep law that is implemented.
Additionally, corresponding creep laws are implemented for both new rheologies.

PR #116, opened 2023-10-12
reworking and adjusting the output table function

Dr3ven:
The old output tables function uses a dict inside a function to find the parameter description/name from the corresponding variable. I want to try to improve this to be more independent of the hard coded dict and descriptions.

Additionally, the tables are adjusted for the new rheologies that were added with #99.

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -1,7 +1,7 @@
 module Tables
 
 using Unidecode
-using GeoParams: AbstractMaterialParam, param_info, LinearViscous, PowerlawViscous, DislocationCreep, DiffusionCreep, CompositeRheology, Parallel, make_tuple
+using GeoParams: AbstractMaterialParam, param_info, LinearViscous, PowerlawViscous, DislocationCreep, DiffusionCreep, GrainBoundarySliding, PeierlsCreep, NonLinearPeierlsCreep, CompositeRheology, Parallel, make_tuple
 using ..Units
 using ..MaterialParameters: MaterialParamsInfo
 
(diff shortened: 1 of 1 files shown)


albert-de-montserrat:
It would be nice to improve `Tables.jl` indeed. Taking a quick look, some suggestions:

1. Get rid of all the `global` variables.
2. There is  a lot of code repetition in `Phase2Dict`. You can avoid all that type-based branching by defining small functions and using multiple dispatch.
3. In general, all  the functions are quite large and monolithic. Splitting them into smaller functions is probably  a good idea when possible. Code will probably more clear and may also lead to code re-usability.

albert-de-montserrat:
Are you still planning to work on this? I am wondering whether it would makes sense to move this out to an external package, something like  `GeoParamsTables.jl`, so we can move out some dependencies that are not really needed when one uses GeoParams in a solver or similar.

Dr3ven:
> Are you still planning to work on this? I am wondering whether it would makes sense to move this out to an external package, something like `GeoParamsTables.jl`, so we can move out some dependencies that are not really needed when one uses GeoParams in a solver or similar.

I do not have a lot of time at the moment, but I will finish this pull request once I have more time again.
Concerning the tables feature, the only output dependency that I use is `Unidecode.jl` as far as I know. So, I don't mind moving the Tables to another package if that helps. I guess I would need to rethink how to code some stuff then, but just do as you thinks suits best for GeoParams.

albert-de-montserrat:
I think it make sense to move it out. It is not really needed inside GP. And I don't think it will change the current code structure, since its already a submodule, just need to update a few lines of code here and there.

boriskaus:
I'm fine with moving this to a separate package

PR #119, opened 2023-10-18
Non-dimensionalization of pre-defined power laws

albert-de-montserrat:
This PR adds an option to directly non-dimensionalize power laws dfined in the database. It's enough to pass the characteristic lenghts when instantiating the rheology struct.

```julia
CharDims = GEO_units(; viscosity=1e19, length=1000km)
SetDiffusionCreep("Dry Anorthite | Rybacki et al. (2006)", CharDims)
```


boriskaus:
Cool - can you add some tests?

boriskaus (approved):

PR #121, opened 2023-10-18
Allow GeoUnits {SI} when non-dimensionalizing predifned rheology

albert-de-montserrat:
Relaxes the type of the characteristic length object to be parameterized as `GeoUnits{GEO}` as well as `GeoUnits{SI}`.


albert-de-montserrat:
@tdurezt could you check if this branch works well for you?

Issue #122, opened 2023-10-20
Problem with having Name in the rheology structs


albert-de-montserrat:
Thanks to @tduretz testing we have realised that having the name of the rheology in the `AbstractCreepLaw` types makes the code type-unstable. For example something like

```julia
function main()
   x = DislocationCreep()
   for i in 1:10
      # do some calculation with x
   end
end
```
will make every iteration allocate. Looks like the whole problem comes from the conversion of `Name` from `String` to `NTuple{N,Char}` because `N` is the length of `Name` and is unknown at compile time. Same problem applies to e.g. `MaterialParams`.

Having `Name` in the object also has another problem, which is that it increases the size of these `struct` to the point that a CUDA kernel will fail to compile if there are many material phases passed into it.

I would propose to just remove `Name` from these objects. I guess an user could simple make a dictionary to map `Phase::Int64` in `MaterialParams` to phase/creep law name if he wants this information.

@boriskaus  any thoughts?

albert-de-montserrat:
For the creeps laws, maybe we could identify them with an integer instead. `Name=0` if they are not from the database, and `Name=x::Int` if they are from the database, then just provide a `get_name`  function to map the integer to the string.

boriskaus:
Hmm, we converted that because it is otherwise not `isbittype`. Is it not possible to encode `N` in the type name as well? I quite like having the name in the struct but if there is no other way we can do it as you suggest

albert-de-montserrat:
A simple workaround may be to just use `Adapt.jl` and leave the name as a `String`. This does the trick

```julia
struct F1{T}
    x::T
    str::String
end
Adapt.@adapt_structure F1

A = F1(1, "potato")

@parallel_indices (i,j) function f(x, A)
    x[i,j] *= 2
    return nothing
end

@parallel (1:2,1:2) f(@rand(2,2), A) # works
```

boriskaus:
That would do it as well; requires us to change this in a few places

PR #126, opened 2023-10-26
[Proposal] Default rheology names to nothing

albert-de-montserrat:
I was too quick with the fix related to the rheology names in #122. 

**Problem** 
`String`s are not isbits and converting them to a isbits representation that is not an array is type unstable:

```julia
str2char(str::String) = ntuple(i->str[i], length(str))
```

```julia
In [18]: @code_warntype str2char(str)
MethodInstance for str2char(::String)
  from str2char(str::String) @ Main c:\Users\albert\Desktop\GeoParams.jl\GeoParams.jl\rot.jl:86
Arguments
  #self#::Core.Const(str2char)
  str::String
Locals
  #36::var"#36#37"{String}
Body::Tuple{Vararg{Char}}
1 ─ %1 = Main.:(var"#36#37")::Core.Const(var"#36#37")
│   %2 = Core.typeof(str)::Core.Const(String)
│   %3 = Core.apply_type(%1, %2)::Core.Const(var"#36#37"{String})
│        (#36 = %new(%3, str))
│   %5 = #36::var"#36#37"{String}
│   %6 = Main.length(str)::Int64
│   %7 = Main.ntuple(%5, %6)::Tuple{Vararg{Char}}
└──      return %7
```

**Solution for GPUs**
The only solution I could find is to use an `Array` instead of a tuple
```julia
xc = [Char(x) for x in x]
```
so that `xc_gpu = CuArray(xc)` works in CUDA kernels (same for ROCArrays).

**Proposal**
Default all the names to `Name=nothing` (including those rheologies in the databases), and let the user fill the name with the appropriate value type. And document this issue somewhere in the docs.

### DIFF
--- a/src/ConstitutiveRelationships.jl
+++ b/src/ConstitutiveRelationships.jl
@@ -8,7 +8,7 @@ using Parameters, LaTeXStrings, Unitful
 using ..Units
 using GeoParams: AbstractMaterialParam, AbstractConstitutiveLaw, AbstractComposite
 import GeoParams: param_info, fastpow, pow_check, nphase, ntuple_idx, @print, @pow
-import GeoParams: second_invariant, second_invariant_staggered, value_and_partial
+import GeoParams: second_invariant, second_invariant_staggered, value_and_partial, printable_name
 using BibTeX
 using ..MaterialParameters: MaterialParamsInfo
 import Base.show
--- a/src/CreepLaw/Data/DislocationCreep.jl
+++ b/src/CreepLaw/Data/DislocationCreep.jl
@@ -187,7 +187,7 @@ function DislocationCreep_data(name::String)
         return DislocationCreep(;
             Name = "Granite | Tirel et al. (2008)",
             n = 3.2NoUnits,
-            A = 1.25e-9MPa^(-3.2) / s,
+            A = 1.25e-9MPa^(-16 // 5) / s,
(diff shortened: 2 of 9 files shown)


boriskaus:
I'm not sure I like this. What's wrong with the previous approach of transforming the String to `Vector{Uint}`? One can perhaps use StaticArrays and encode the length?

albert-de-montserrat:
`Vector{Uint}` is fine, but needs to be moved to a GPU Array to work with them. Meaning that we need to add CUDA/AMDGPU as dependencies if this needs to be done internally.

The original approach was to convert them to a `NTuple{N, Char}` and I'm afraid StaticArrays won't help here. The problem is the same, the length of the string is not known at compile time.

boriskaus:
> `Vector{Uint}` is fine, but needs to be moved to a GPU Array to work with them. Meaning that we need to add CUDA/AMDGPU as dependencies if this needs to be done internally.
> 
> The original approach was to convert them to a `NTuple{N, Char}` and I'm afraid StaticArrays won't help here. The problem is the same, the length of the string is not known at compile time.

Can we not simply add `N` to the type definition of the `structs`? The strings for the different structs is known when we construct them.

albert-de-montserrat:
☝️ does the trick. I close this PR and revert the changes back to its original form.

PR #130, opened 2023-10-30
compute melt fraction with PhaseRatios

aelligp:
This provides the same framework for melt fraction as for density and utilises `compute_param_times_frac`.


albert-de-montserrat (approved):
Great thanks for the contribution! Looks good to go.

aelligp:
Can you think of any other parameters that need the PhaseRatio computation? Viscosity already has it.

albert-de-montserrat:
We can add them as we need them :)  I will merge this

PR #136, opened 2023-12-01
set default values for G and nu in SetConstantElasticity()

Dr3ven:
Using nothing for `G` and `ν` makes `SetConstantElasticity()` without any input arguments produce `NaN`s. To fix that default values without Units are set.


albert-de-montserrat (approved):
Looks good to me thanks!

PR #139, opened 2023-12-04
Test allocations

albert-de-montserrat:
After the type stability improvements, we can uncomment the old `@test num_alloc <= 32` and replace them by `@test num_alloc == 0`


albert-de-montserrat:
I will put this on hold, AllocChecks.jl doesn't seem to be very robust in Windows yet.

PR #146, opened 2024-01-02
Fix instabilities in `compute_units`

albert-de-montserrat:
Fixes the type instabilities in  `compute_units` which had an impact in the non/dimensionalization functions.

```julia
CharUnits_GEO = GEO_units(; viscosity=1e19, length=1000km)
```

Main branch
```julia-repl
julia> @btime nondimensionalize($(10cm / yr, CharUnits_GEO)...)
  4.827 μs (66 allocations: 2.69 KiB)
```

This PR
```julia-repl
julia> @btime nondimensionalize($(10cm / yr, CharUnits_GEO)...)
  161.724 ns (5 allocations: 80 bytes)
```

The remaining allocations are because `GeoUnits` is not parameterized ([here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/a21703cb285cf07479e3023677d21827b67f9f29/src/Units.jl#L312)). I'm not sure it's worth the effort though.

### DIFF
--- a/src/Units.jl
+++ b/src/Units.jl
@@ -337,11 +337,11 @@ end
     #Luminosity
 
     # Derived units
-    N = kg * m / s^2        # Newton
+    N = kg * m / s^2      # Newton
     J = N * m             # Joule
     W = J / s             # Watt
-    area = m^2             # area   in SI units for material parameter scaling
-    volume = m^3             # volume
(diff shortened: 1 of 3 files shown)


boriskaus (approved):
fantastic! I agree it is not worth hunting the rest down (that is usually only done once per simulation)

PR #148, opened 2024-01-08
Fixes bug id 3D shear heating kernel

albert-de-montserrat:
Fixes a bug when a tuple with the tensors in Voigt notation was passed into the 3D shear heating kernel.

This PR also expands the shear heating test coverage.


albert-de-montserrat (commented):
@aellig this fixes the bug you encountered

PR #150, opened 2024-01-10
Bump Makie to v0.20 releases

aelligp:
- additional workflow could include [CompatHelper.jl](https://github.com/JuliaRegistries/CompatHelper.jl)


albert-de-montserrat:
Is `size` completely equivalent to  the old `resolution`? or need different numbers for the same size figure?

aelligp:
From what I've read it is just a more neutral term for the same thing as resolution [see here](https://github.com/MakieOrg/Makie.jl/pull/3343#issue-1971756373).

albert-de-montserrat (approved):

PR #152, opened 2024-01-22
Plastic strain softening

albert-de-montserrat:
Adds (linear) strain softening to cohesion and/or friction angle as a function of accumulated plastic strain.


boriskaus:
looks good to me

albert-de-montserrat:
Partially closes the first point  in  [this issue](https://github.com/PTsolvers/JustRelax.jl/issues/88)

PR #158, opened 2024-01-26
Non linear softening

albert-de-montserrat:
Adds non linear softening from [Thibault et al 21](https://agupubs.onlinelibrary.wiley.com/doi/pdfdirect/10.1029/2021GC009675)

### DIFF
--- a/src/ConstitutiveRelationships.jl
+++ b/src/ConstitutiveRelationships.jl
@@ -62,7 +62,8 @@ export param_info,
     iselastic,
     AbstractSoftening,
     NoSoftening,
-    LinearSoftening
+    LinearSoftening,
+    NoLinearSoftening
 
 # add methods programmatically 
 for myType in (:LinearViscous, :DiffusionCreep, :DislocationCreep, :ConstantElasticity, :DruckerPrager, :ArrheniusType, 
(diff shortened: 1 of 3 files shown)


boriskaus:
why is this called `NoLinearSoftening` instead of `NonlinearSoftening`? Typo?

src/Softening/Softening.jl
```diff
@@ -29,3 +30,22 @@ LinearSoftening(min_max_values::NTuple{2, T}, lo_hi::NTuple{2, T}) where T = Lin
     
     return fma((one(T) - softening_var), softening.slope, softening.min_value)
 end
+
+## Non linear softening 
+# (Thibault et al 2021; https://agupubs.onlinelibrary.wiley.com/doi/pdfdirect/10.1029/2021GC009675)
```
boriskaus, line 35: Duretz et al.

src/ConstitutiveRelationships.jl
```diff
@@ -62,7 +62,8 @@ export param_info,
     iselastic,
     AbstractSoftening,
     NoSoftening,
-    LinearSoftening
+    LinearSoftening,
+    NoLinearSoftening
```
boriskaus, line 66: would be better to call it `NonlinearSoftening`

albert-de-montserrat:
> why is this called `NoLinearSoftening` instead of `NonlinearSoftening`? Typo?

it is indeed a typo... i'm patching it

PR #163, opened 2024-01-31
Add latent heat by modifying heat capacity 

boriskaus:
There are two ways to implement latent heat in the energy equation: explicitly by adding it as a source term or implicitly. The explicit manner was already available `GeoParams`; this PR adds the implicit method which implements it by modifying the thermal heat capacity (making it temperature-dependent).

The are several texts on how to implement this; since some lessons learned seemed to have gone forgotten, I recently summarised this [here](https://pubs.geoscienceworld.org/gsa/geosphere/article/19/4/1006/624461/Zircon-age-spectra-to-quantify-magma-evolution).

In short, the energy equation (w/out advection) is:
```math
\rho c_p \frac{\partial T}{\partial t}  ) = \frac{\partial }{\partial x_i} \left( k   \frac{\partial T}{\partial x_i}  \right)  + H_s + \rho Q_L\frac{\partial \phi_s}{\partial t}  
```
where $H_s$ are heat sources, $\phi_s=(1-\phi)$ is the solid fraction, $\phi$ the melt fraction, $k$ thermal conductivity, $\rho$ density, $c_p$ heat capacity and $Q_L$ the latent heat. 

Using the chain rule we can rewrite the last term as
```math
\rho c_p \frac{\partial T}{\partial t}  =\frac{\partial }{\partial x_i} \left( k   \frac{\partial T}{\partial x_i}  \right)  + H_s + \rho Q_L\frac{\partial \phi_s}{\partial T}\frac{\partial T}{\partial t} . 
```
Since 
```math
\frac{\partial \phi_s}{\partial T}   = -\frac{\partial \phi}{\partial T},  
```
```math
\rho c_p \frac{\partial T}{\partial t}  =\frac{\partial }{\partial x_i} \left( k   \frac{\partial T}{\partial x_i}  \right)  + H_s - \rho Q_L\frac{\partial \phi}{\partial T}\frac{\partial T}{\partial t},
```
we obtain:
```math
\rho \left( c_p + Q_L\frac{\partial \phi}{\partial T} \right) \frac{\partial T}{\partial t}  =\frac{\partial }{\partial x_i} \left( k   \frac{\partial T}{\partial x_i}  \right)  + H_s. 
```

From this it is clear that we can rewrite the original equation as:
```math
\rho c_p^{\textrm{new}} \frac{\partial T}{\partial t}  = \frac{\partial }{\partial x_i} \left( k   \frac{\partial T}{\partial x_i}  \right)  + H_s,
```
with
```math
c_p^{\textrm{new}} = c_p + Q_L\frac{\partial \phi}{\partial T}.
```

So we can implement latent heat my modifying the heat capacity. This formulation is numerically more stable than by adding it as a source term. A point to remember, though, is that it is crucial that  $\frac{\partial \phi}{\partial T}$ does not have discontinuities. Many of the parameterisations in use in the petrology/volcanological community do have jumps around $\phi=0$ or $\phi=1$ which results in non-converging  numerical solutions.

With the current PR, you can add latent heat as:
```Julia
julia> x1  = Latent_HeatCapacity(Q_L=500kJ/kg)
Latent heat through modifying heat capacity: Cp = Cp + dϕdT*500.0 kJ kg⁻¹·⁰; Cp=Constant heat capacity: cp=1050.0 J kg⁻¹·⁰ K⁻¹·⁰
```
Note that you can use any of the available heat capacity functions, including temperature-dependent ones:
```Julia
julia> x2 = Latent_HeatCapacity(Cp=T_HeatCapacity_Whittington())
Latent heat through modifying heat capacity: Cp = Cp + dϕdT*400.0 kJ kg⁻¹·⁰; Cp=T-dependent heat capacity following Whittington et al. (2009) for average crust. 
```

### DIFF
--- a/docs/src/man/latentheat.md
+++ b/docs/src/man/latentheat.md
@@ -1,12 +1,23 @@
 # Latent heat
 
 # Methods
-Latent heat (of crystallisation) is defined as 
+Latent heat (of crystallisation) can be implemented as a source term (usually numerically not very stable):
 ```@docs
-GeoParams.MaterialParameters.LatentHeat.ConstantLatentHeat
+ConstantLatentHeat
 ```
+Alternatively, you can implement it by modifying the heat capacity, which is often numerically better.
+```@docs
+Latent_HeatCapacity
+```
+
 # Computational routines
-To compute, use this:
+To compute with the source term, use this:
 ```@docs
 GeoParams.MaterialParameters.LatentHeat.compute_latent_heat
+```
+
+If you modify the heat capacity, you simply use this in your thermal computations:
+To compute with the source term, use this:
+```@docs
+compute_heatcapacity
 ```
--- a/src/Energy/Conductivity.jl
+++ b/src/Energy/Conductivity.jl
@@ -23,7 +23,6 @@ export compute_conductivity,                       # calculation routines
     Set_TP_Conductivity                         # Routine to set pre-defined parameters
 
 include("../Computations.jl")
-#include("../Utils.jl")
 
 # Constant Conductivity -------------------------------------------------------
 """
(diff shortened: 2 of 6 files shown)


aelligp (approved):
looks good to me

PR #172, opened 2024-02-12
Extractor methods

albert-de-montserrat:
Sometimes I need to extract parameters from different types which they may be inside tuples of different phases or whatever. This PR adds the `@extractors` macro to automatically generate functions to extract variables from a `struct`. An example for the `AbstractDensity` case [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/04bcbfdc030b9aec5a6c5564ca6c0ff00fdac181/src/Density/Density.jl#L375-L379)
```
for type in (ConstantDensity, PT_Density, Compressible_Density, T_Density, MeltDependent_Density)
    @extractors(type, :Density)
end
```

generates e.g. `get_α(x::PT_Density) = x.α.val` and the appropriate methods for `AbstractMaterialParamsStruct` and `Tuple{N, AbstractMaterialParamsStruct}`. 

In action:
```julia-repl
In [1]: using GeoParams

In [1]: x = PT_Density()
P/T-dependent density: ρ0=2900.0 kg m^-3.0, α=3.0e-5 K^-1.0, β=1.0e-9 Pa^-1.0, T0=0.0 °C, P0=0.0 MPa

In [2]: GeoParams.get_α(x)
3.0e-5
```

### DIFF
--- a/src/ConstitutiveRelationships.jl
+++ b/src/ConstitutiveRelationships.jl
@@ -8,7 +8,7 @@ using Parameters, LaTeXStrings, Unitful
 using ..Units
 using GeoParams: AbstractMaterialParam, AbstractConstitutiveLaw, AbstractComposite
 import GeoParams: param_info, fastpow, pow_check, nphase, ntuple_idx, @print, @pow, str2tuple, uint2str
-import GeoParams: second_invariant, second_invariant_staggered, value_and_partial
+import GeoParams: second_invariant, second_invariant_staggered, value_and_partial, @extractors, add_extractor_functions
 using BibTeX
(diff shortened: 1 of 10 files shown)


boriskaus:
That’s very helpful indeed - I guess it can be used to remove quite a bit of code?

Would be great if you can add a test.

albert-de-montserrat:
Well right now we don't have many of these kind of functions so it doesn't really reduce a lot of code, but it reduces the amount of effort to add them :) Still need to add this to other fields apart from density, then ill add the tests

boriskaus (approved):

Issue #170, opened 2024-02-08
Bug in `T_HeatCapacity_Whittington` ?


albert-de-montserrat:
Given 

```julia
 CharUnits_GEO = GEO_units(; viscosity=1e19, length=10km)
 x_D1 =Latent_HeatCapacity(Cp=T_HeatCapacity_Whittington())
 x_ND1 = nondimensionalize(x_D1, CharUnits_GEO)
 
 dϕdT = 0.1
 args = (dϕdT=dϕdT, T=10)
 ```

It then yields
```julia-repl
In [9]:  Cp = compute_heatcapacity(x_D1, args)
-224505.23852466405
```
@boriskaus  I guess there's a bug somewhere

albert-de-montserrat:
My bad, I think it is correct... T has to be in K :weary:

boriskaus:
indeed K it should be. We should probably clarify that in the help.

It should actually also work if we nondimensionalize everything and redimensionalize it after computation. Would perhaps be good to add a test to verify this, given that you now spend some time looking into this

aelligp:
I'll address this in #169

albert-de-montserrat:
Am I doing something wrong?

```julia
CharUnits_GEO = GEO_units(; viscosity=1e19, length=10km)
x_D =Latent_HeatCapacity(Q_L=500kJ/kg)
x_ND = nondimensionalize(x_D, CharUnits_GEO)

dϕdT = 0.1
args = (;dϕdT)
Cp = compute_heatcapacity(x_D, args)
Cp_nd = compute_heatcapacity(x_ND, args)
Cp_nd0 = nondimensionalize(Cp*(J / kg / K), CharUnits_GEO)
```

gives

```julia-repl
In [24]: dimensionalize(Cp_nd, J / kg / K, CharUnits_GEO)
1089.2726701488434 J kg^-1.0 K^-1.0
In [25]: dimensionalize(Cp_nd0, J / kg / K, CharUnits_GEO)
1100.0 J kg^-1.0 K^-1.0
```

aelligp:
this looks like a roundoff error as: 
```
julia> CharUnits_GEO.heatcapacity
7.854534029768683e-20 m²·⁰ K⁻¹·⁰ s⁻²·⁰
```

albert-de-montserrat:
Shouldn't 
```julia
[...]

PR #173, opened 2024-02-13
Set units of type kwargs to SI

albert-de-montserrat:
Fixes the problem raised by #170 related non-SI units in kwargs of different types.


src/CreepLaw/NonLinearPeierlsCreep.jl
```diff
@@ -53,8 +53,8 @@ struct NonLinearPeierlsCreep{T,N,U1,U2,U3,U4,U5} <: AbstractCreepLaw{T}
         q=1.0NoUnits,
         o=0.5NoUnits,
         TauP=8.5e9Pa,
-        A=5.7e11MPa^(-2) * s^(-1),
-        E=476.0kJ / mol,
+        A=0.57Pa^(-2) * s^(-1),
```
aelligp, line 56: 0.57Pa isn't 5.7e11
```suggestion
        A=5.7e17Pa^(-2) * s^(-1),
```

aelligp (changes requested):

src/CreepLaw/NonLinearPeierlsCreep.jl
boriskaus: As per README @ GeoParams:
```Julia
julia> using GeoParams;
julia> A=5.7e11MPa^(-2) * s^(-1)
5.7e11 MPa⁻²·⁰ s⁻¹·⁰
julia> uconvert(Pa^-2*s^-1, A)
0.57 Pa⁻²·⁰ s⁻¹·⁰
```
I guess every geodynamicists must have fallen into this trap at least once.

aelligp (approved):

PR #182, opened 2024-03-02
Create DocPreviewCleanup.yml

ranocha:
When you have many PRs from branches of the repo itself (instead of from forks), Documenter.jl will build quite a lot of previews for the PRs. These previews may bloat your `gh-pages` branch. To avoid that, you can use a GitHub action like the one we're using in the Trixi.jl framework.

See also https://github.com/JuliaGeodynamics/GeophysicalModelGenerator.jl/pull/74


aelligp (approved):
Thanks

PR #183, opened 2024-03-03
Viscosity methods for linear rheologies

albert-de-montserrat:
Introduces `compute_viscosity` function for `LinearViscous` and `ConstantElasticity` objects:

```julia
η0  = 3
dt  = 4
args = (; dt=dt)
G        = 1
el       = SetConstantElasticity(; G=G, ν=0.5)
creep    = LinearViscous(; η=η0)
rheology = SetMaterialParams(; CompositeRheology=CompositeRheology((creep, el)))

@assert compute_viscosity(el, args) == G * dt  # true
@assert compute_viscosity(creep, args) == η0 # true
@assert compute_viscosity(rheology, args) == 1/(1 / η0 + 1 / G / dt) # true
```

`compute_elasticviscosity` can be used to compute `G * dt`
```julia
rheologies2 = (
    SetMaterialParams(; 
        CompositeRheology=CompositeRheology((LinearViscous(; η=1), el))
    ),
    SetMaterialParams(; 
        CompositeRheology=CompositeRheology((LinearViscous(; η=2), el))
    )
)

@assert compute_elasticviscosity(rheologies2, 1, args) == el.G.val * dt
@assert compute_elasticviscosity(rheologies2, 2, args) == el.G.val * dt
@assert compute_elasticviscosity(rheologies2, phase_ratio, args) == el.G.val * dt
```

All the new functions work with all the GeoParams machinery; more examples in the [tests](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/791abc5bdcb908f296b7ca1e58c25ffd02248690/test/test_Viscosity.jl#L29)

Also fixes the problem when passing zero stress or strain to  `compute_viscosity_εII` and `compute_viscosity_τII`

```julia-repl
julia> compute_viscosity_τII(creep, 0.0, args)
3.0

julia> compute_viscosity_τII(rheology, 0.0, args)
3.0
```

### DIFF
--- a/src/GeoParams.jl
+++ b/src/GeoParams.jl
@@ -263,7 +263,8 @@ export compute_viscosity_εII,
     compute_elastoviscosity,
     compute_elastoviscosity_εII,
     compute_elastoviscosity_τII,
-    compute_viscosity
+    compute_viscosity,
+    compute_elasticviscosity
 
 # Gravitational Acceleration
 using .MaterialParameters.GravitationalAcceleration
--- a/src/GeoParams.jl
+++ b/src/GeoParams.jl
@@ -262,7 +262,8 @@ export compute_viscosity_εII,
     compute_viscosity_τII,
     compute_elastoviscosity,
     compute_elastoviscosity_εII,
-    compute_elastoviscosity_τII
+    compute_elastoviscosity_τII,
+    compute_viscosity
 
 # Gravitational Acceleration
 using .MaterialParameters.GravitationalAcceleration
--- a/src/Viscosity/Viscosity.jl
+++ b/src/Viscosity/Viscosity.jl
@@ -17,7 +17,7 @@ Compute effective viscosity given a 2nd invariant of the deviatoric strain rate 
 end
 
 @inline compute_viscosity_εII(v::LinearViscous, εII, args) = v.η.val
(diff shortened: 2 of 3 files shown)


tduretz:
Cool, great that the issue with passing zero stress is gone. Is this syntax essential `el.G.val`? 
What else that a `val` may be stored in `el.G`? Could be nicer to have simply `el.G`.

boriskaus:
`el.G` is a `GeoUnit` as we somehow need info about the dimensions when we nondimensionalize it. It has a `val` for the value, a `unit` and a flag indicating whether it is dimensional or not.
I guess we could think about overloading this when we multiply it with other numbers such that we don’t have to indicate `val` all the time. We do provide the `NumValue` function for this, but I guess you will like that even less

albert-de-montserrat (commented):
> I guess we could think about overloading this when we multiply it with other numbers such that we don’t have to indicate `val` all the time. 

Actually, I forgot that we alrady do [this](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/90dd56fbd600fb0929ddeb08ab9f394bc3a7c51a/src/Units.jl#L236)

tduretz:
Should I wait for a green CI badge to positively review? ;)

albert-de-montserrat:
Its only nightly for some reason... you can ignore that :)

tduretz (approved):

PR #192, opened 2024-04-08
Vector density, melt fraction and heat capacity 

boriskaus:
This PR allows specifying vectors with `Density`, `HeatCapacity` and `MeltFraction` data and retrieving them with indices.


albert-de-montserrat (approved):
Looks good to me!

PR #199, opened 2024-05-13
Store rheology and phase names as pointers

albert-de-montserrat:
I am still facing problems on some CUDA cards (in particular on the Daint), where if we have several material phases in a `tuple` of `MaterialParams`, it won't compile because the size of this object exceeds what the card allows.  The main contributor to the size of `MaterialParams` is the name, which is currently stored as a `NTuple{100, Uint8}`.

The resulting object is still `isbits(MatParam) === true` and works on GPUs.
This PR changes this approach and uses a pointer to a [interned ](https://github.com/JuliaString/InternedStrings.jl) `String`, which is significantly lighter:

```julia
julia> using InternedStrings

julia> str = "Viscous Matrix"
"Viscous Matrix"

julia> sizeof(pointer(intern(str)))
8

julia> sizeof(ntuple(_ -> rand(UInt8), Val(100)))
100
```

And in a simple test case
```julia
using GeoParams

MatParam = (
    SetMaterialParams(;
        Name="Viscous Matrix",
    ),
    SetMaterialParams(;
        Name="Viscous Sinker"
    ),
    SetMaterialParams(;
        Name="Viscous Bottom",
    ),
)
```

```julia
julia> sizeof(MatParam) # this branch
72 

julia> sizeof(MatParam) # main branch
360
```


boriskaus (approved):

albert-de-montserrat:
Fixes my problems on the Daint :)

PR #202, opened 2024-06-14
Add perturbation kwarg to cohesion

albert-de-montserrat:
Adds a `pertubation_C` keyword for `DruckerPrager` and `DruckerPrager_regularized` so that the cohesion can be perturbation such that `C_perturbed = C * pertubation_C`


aelligp (approved):
Looks good - thank you

PR #205, opened 2024-06-28
Bump GLMakie

albert-de-montserrat:
Might fix CI on 1.11


boriskaus:
It seems that the issue is that the X11 display is missing if running this on the CI system. Perhaps they changed the way to switch that off - may be worthwhile to check how they do the CI in Makie

luraess:
Wouldn't it be more reliable to have Cairo as "default" backend here? GL requires an OpenGL context and may fail on headless servers in general... (ignore this comment if this si somehow already addressed).

aelligp:
This has been addressed by using the base extension `using Makie` in #254

PR #224, opened 2024-10-17
Update blank.yml

aelligp:
Addresses the CI on macOS issue reported by @boriskaus


boriskaus (approved):

boriskaus:
Looks like we have to specify julia 1.9 as lower bound. I’m fine with that

PR #226, opened 2024-10-19
Add power law strain/stress functions

albert-de-montserrat:
Power law creep had a `struct` already defined in `src/CreepLaw.jl`. However, there were no methods associated to it.


boriskaus (approved):
Thanks - indeed long overdue

PR #233, opened 2024-11-26
New functionality to nondimensionalisation

aelligp:
This PR adapts the (non-)dimensionalisation functions to also work if no characteristic unit is provided. It will return the original values without the units and adds a warning that char. units is not provided. 
This could be useful if a script is usually non-dimensional but one would like to try with dimensional numbers.
E.g.:
```julia-repl
julia> nondimensionalize(10cm/yr,nothing)
┌ Warning: The input parameter is not being nondimensionalized, as no characteristic units are given
└ @ GeoParams.Units ~/Documents/PhD/GeoParams.jl/src/Units.jl:610
10

julia> nondimensionalize(T, nothing)
┌ Warning: The input parameter is not being nondimensionalized, as no characteristic units are given
└ @ GeoParams.Units ~/Documents/PhD/GeoParams.jl/src/Units.jl:642
3×3 Matrix{Int64}:
 10  10  10
 10  10  10
 10  10  10
 ```

### DIFF
--- a/src/Units.jl
+++ b/src/Units.jl
@@ -16,7 +16,7 @@ using GeoParams:
     AbstractPhaseDiagramsStruct,
     PerpleX_LaMEM_Diagram
 
-# Define additional units that are useful in geodynamics 
+# Define additional units that are useful in geodynamics
 @unit Myrs "Myrs" MillionYears 1000000u"yr" false
 
 function __init__()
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by boriskaus — APPROVED** [review]
thanks!

File: src/Units.jl [hunk 1]
```diff
@@ -570,8 +570,11 @@ julia> upreferred(A)
 3.1574795718518295e-20 m³·⁰⁵ s⁵·¹ kg⁻³·⁰⁵
 ```
 """
-function nondimensionalize(param::GeoUnit{T,U}, g::GeoUnits{TYPE}) where {T,U,TYPE}
-    if param.isdimensional
+function nondimensionalize(param::GeoUnit{T,U}, g::Union{GeoUnits{TYPE}, Nothing}) where {T,U,TYPE}
+    if isnothing(g)
+        @warn("The input parameter is not being nondimensionalized, as no characteristic units are given")
+        return param_ND = param
+    elseif param.isdimensional
```
albert-de-montserrat, line 577: This is better done using multiple dispatch. You can leave the original method as it is and add a new one

```julia
function nondimensionalize(param::GeoUnit, g::Nothing)
        @warn("The input parameter is not being nondimensionalized, as no characteristic units are given")
        return param
end
```

albert-de-montserrat (approved):
Great thanks!

PR #234, opened 2024-11-26
Add a downstream pipeline for testing

aelligp:
In this file we can add repos that are directly dependent on GeoParams and let the CI tests for the specified pkg run


albert-de-montserrat (approved):
This is great thanks!

PR #235, opened 2024-11-27
AD type management and implementation (part I/X)

jfrasunk:
This PR adds some of the AD things needed for adjoint etc done a while ago by @boriskaus and @albert-de-montserrat


src/Permeability/Permeability.jl
```diff
@@ -25,11 +25,11 @@ export compute_permeability,     # calculation routines
 
 # Define "empty" computational routines in case nothing is defined
 function compute_permeability!(
-    k::_T, s::No_MaterialParam{_T}; ϕ::_T=zero(_T)
-) where {_T}
+    k, s::No_MaterialParam{_T}; ϕ::_T1=zero(_T1)
```
albert-de-montserrat, line 28: This will not work, because the compiler does not know what `_T1` is unless `ϕ` is explicitly passed as a keyword argument. Example:
```julia
julia> foo(; x::T=zero(T)) where T = x
foo (generic function with 1 method)

julia> foo()
ERROR: UndefVarError: `T` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] foo()
   @ Main .\REPL[19]:1
 [2] top-level scope
   @ REPL[20]:1
```

albert-de-montserrat (changes requested):

src/Permeability/Permeability.jl
aelligp: So you're saying that this also wouldn't work? and every zero() should be changed to 0.0?

```julia
function compute_density!(
     rho, s::No_MaterialParam{_T}; P::_T1=zero(_T1), T::_T2=zero(_T2)
 ) where {_T, _T1, _T2}
     return zero(_T)
 end
 function compute_density(s::No_MaterialParam{_T}; P::_T1=zero(_T1), T::_T2=zero(_T2)) where {_T,_T1,_T2}
 ```

albert-de-montserrat: Yep, that won't work if you don't pass `P` and `T`. Better make it ` P::_T1=0e0, T::_T2=0e0`

albert-de-montserrat: You should change every `zero(T)` where `T` is not known/inferred from somewhere else.

albert-de-montserrat (approved):

Issue #243, opened 2025-01-22
Myrs should be Myr


Iddingsite:
I believe Myrs as a unit should be replaced by Myr, which is the commonly used abbreviation (see for instance https://en.wikipedia.org/wiki/Million_years_ago), which caused me some confusion the first time. All the other units are using conventional abbreviations and I think it would be nice to be consistent there.

That's something very easy to fix in the code but it would be a breaking change as it is part of the exposed API, so I don't know if that is a change you want to do now, but for me it should be ultimately corrected.

boriskaus:
I'm fine with deprecating `Myrs`. Perhaps introduce `Myr` and put a depreciation warning for some time that this will be removed in the future. I believe there is a Julia command for that.

luraess:
Yes it has

Iddingsite:
Solved by #246.

PR #244, opened 2025-01-22
replace Myrs by Myr

Iddingsite:
Fix https://github.com/JuliaGeodynamics/GeoParams.jl/issues/243.

Not super satisfying on the warning side, because as Myrs is defined as a constant, it runs the warning only the first time `using Geoparams` is called but not when `Myrs` is used. 

If someone has a better way, tell me!


luraess:
IDK if one could use some deprecation mechanism as in e.g. https://github.com/JuliaGPU/CUDA.jl/blob/master/src/deprecated.jl ?

aelligp:
As far as I've understood it is only applicable if it's a function. But not with a unit

boriskaus (approved):
Thanks a lot!

PR #241, opened 2025-01-15
Adding setup for chemical diffusion data

Iddingsite:
Hi,

It has been a while since I wanted a package to compile chemical diffusion data from mineral and phases with some routines and I think it would be nice to host that directly in GeoParams.jl, to take advantage of everything already in there. That would also expend the usage of this package for smaller scale, which can be nice to have a broader community.

As such, this pull-request attempts to do so, and is still WIP. My idea is to use the same structure of CreepLaw more or less, with a folder Data containing all the data from the litterature, and routines to calculate diffusion coefficients. 

I think the structure I want is there with some tests (which needs to be extended), but they are some details that may be worth discussing at some point. Inputs are welcome!

PR #248, opened 2025-01-24
Add more data for chemical diffusion and some refinements

Iddingsite:
Hey,

This is a follow-up PR on #241, with some data from the literature (Major et REE in garnet, Traces in rhyolitic melt) and also some tweaking on the original structure from what I learn collecting these data. Everything is benchmarked. Lots of lines of code, but it is mainly Data from papers.

Also update Static and ForwardDiff because that was blocking compatibility with some packages.


aelligp (approved):
Looks good to me, thanks for the contribution

PR #252, opened 2025-02-06
add more Rutile chemical diffusion data

monalueder:
Hi,

I have added new chemical diffusion data for most of the elements of Rutile. I have also updated the documentation and added tests.

PR #254, opened 2025-02-08
Adding plot for diffusion coef, improving docs, use Makie and small fixes

Iddingsite:
Hi, 

This is a pull-request doing a few different things and fixes:

- Make the plotting extension dependent on Makie and not GLMakie so we can use other backends (i.e. CairoMakie), which was not possible before. This is not breaking as CairoMakie or GLMakie call Makie behind the hood.
- Add latex equations support with Documenter.jl via MathJax (default is Katex and is limited in some things, like numbering equations for instance and load Physics package)
- Fix Documenter.jl link with the extension modules, which is currently broken
- Add a plotting function to make arrhenius plot for diffusion coefficient (`PlotDiffusionCoefArrhenius`) to make this kind of plot: 
ChemicalDiffusion_garnetdiffusion
- fixes a few mistakes in the chemical diffusion coefs introduced in #252
- improves docs for the chemical diffusion section


boriskaus:
That’s very cool - thanks! If you want you can also switch off testing for nightly; this causes a lot of errors

boriskaus (approved):

PR #263, opened 2025-03-06
Added upgrade_GeoUnits() for better compatibility with old .jld2 files

ArneSpang:


boriskaus (changes requested):
perhaps add a docstring explaining what it does and make sure it is added to the documentation

ArneSpang:
> perhaps add a docstring explaining what it does and make sure it is added to the documentation

I have not done this in a while. I added the docstring but where do I need to add the function that it is added to documentation?

boriskaus:
that's [here](https://github.com/JuliaGeodynamics/GeoParams.jl/edit/main/docs/src/man/listfunctions.md) but from the looks of it it should be automatically added

albert-de-montserrat (approved):

PR #266, opened 2025-03-18
Update TAS  Diagram plotting

aelligp:
Updates the `Plot_TAS_diagram` to plot the TAS polygons and compositions on top. Composition has to be an Array. 

Screenshot 2025-03-18 at 14 35 11

### DIFF
--- a/src/Plotting/Plotting.jl
+++ b/src/Plotting/Plotting.jl
@@ -839,11 +839,26 @@ end
 
 
 """
-	plt = Plot_TAS_diagram(; displayLabel=true,size=(1500,1500), fontsize=18)
+	plt = Plot_TAS_diagram(point; displayLabel=true,size=(1500,1500), fontsize=18)
 
 Creates a TAS diagram plot
+
+## Input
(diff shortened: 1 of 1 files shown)


src/Plotting/Plotting.jl
```diff
@@ -839,11 +839,27 @@ end
...
 Creates a TAS diagram plot
+
+## Input
+- `point::AbstractArray`: point to plot on the TAS diagram (SiO2, Na2O+K2O)
+
+## Optional parameters
+- `displayLabel::Bool`: display the label of the lithology
+- `size::Tuple`: size of the plot
+- `fontsize::Int`: fontsize of the plot
+- `colormp::Symbol`: colormap to use
```
boriskaus, line 853: Why not call it `colormap`?

boriskaus (changes requested):
Thanks - lgtm apart from a very minor issue

src/Plotting/Plotting.jl
aelligp: This was mainly to be consistent with the argument `fontsz` in the kwargs. It works even with fontsize and colormap so I changed it.

PR #268, opened 2025-03-20
rename DensityX to Melt_DensityX

aelligp:
For consistency DensityX is renamed to Melt_DensityX as it is only applicable for liquids. 

## Additionally:
- add precomputed `α` and `β` to struct


src/Density/Density.jl
```diff
@@ -457,6 +459,8 @@ struct DensityX{T, T1, T2, T3, T4, T5, T6, U, U1, U2, U3} <: AbstractDensity{T}
             norm_MP = (0.511, 0.012, 0.09, 0.083, 0.117, 0.123, 0.028, 0.001, 0.034) .* NoUnits,
             P0 = 1.0e5Pa,
             ρ0 = 2900.0kg / m^3,
+            α = 2.266e-4 / K,
+            β = -6.319e-11 / Pa,
```
albert-de-montserrat, line 463: Are we sure about the sign of `β`?

boriskaus: but isn't pressure in typical geodynamic models positive with increasing depth?  Doesn't that imply that the density of materials decreases with increasing P?

aelligp: There was a sign mistake as:
 ```math
\beta  = -\frac{1}{V}\frac{\partial V}{\partial P}
```

The one order of magnitude difference in thermal expansivity is consistent with literature. see e.g. [table 3](https://ajsonline.org/article/115477-thermal-expansion-of-metal-oxides-in-silicate-melts-a-simple-theoretical-analysis)

albert-de-montserrat (approved):

PR #270, opened 2025-03-22
multicomponent diffusion in basaltic melt

Iddingsite:
Ok, this PR is a bit big. And I think it requires a proper review. It should also be merged after #269 because this PR is starting from this branch.

It is implementing a way to compute a multicomponent matrix diffusion for basaltic melt for 8 components (SiO2–TiO2–Al2O3–FeO–MgO–CaO–Na2O–K2O), using the approach of Guo and Zhan (2020). 

As it differs a bit to what the structure `DiffusionData`does, I had to implement a custom one, named `MeltMulticompDiffusionData`, so I could also do some multiple dispatch. As it involves matrices, I used StaticArrays so that it doesn't allocate. 

So:

```julia
using GeoParams

melt_multicomponent = Melt.Melt_multicomponent_major_Guo2020_SiO2_basaltic
melt_multicomponent = SetMulticompChemicalDiffusion(melt_multicomponent)

@btime compute_D($melt_multicomponent, T = $(1500C));
```
gives:  1.600 μs (0 allocations: 0 bytes).

and where the output is:

```julia
7×7 StaticArraysCore.SMatrix{7, 7, Quantity{Float64, 𝐋²·⁰ 𝐓⁻¹·⁰, Unitful.FreeUnits{(m²·⁰, s⁻¹·⁰), 𝐋²·⁰ 𝐓⁻¹·⁰, nothing}}, 49} with indices SOneTo(7)×SOneTo(7):
  1.88201e-11 m²·⁰ s⁻¹·⁰   2.89112e-12 m²·⁰ s⁻¹·⁰  …   7.97547e-11 m²·⁰ s⁻¹·⁰   3.53691e-11 m²·⁰ s⁻¹·⁰
 -1.76712e-12 m²·⁰ s⁻¹·⁰   1.8085e-11 m²·⁰ s⁻¹·⁰      -8.35494e-12 m²·⁰ s⁻¹·⁰  -5.85127e-12 m²·⁰ s⁻¹·⁰
 -1.22005e-11 m²·⁰ s⁻¹·⁰   7.74996e-12 m²·⁰ s⁻¹·⁰     -3.71793e-11 m²·⁰ s⁻¹·⁰  -2.1956e-11 m²·⁰ s⁻¹·⁰
  5.62722e-13 m²·⁰ s⁻¹·⁰   1.9547e-11 m²·⁰ s⁻¹·⁰       2.07793e-10 m²·⁰ s⁻¹·⁰   3.78884e-11 m²·⁰ s⁻¹·⁰
  2.72967e-13 m²·⁰ s⁻¹·⁰  -8.01234e-12 m²·⁰ s⁻¹·⁰     -5.64054e-11 m²·⁰ s⁻¹·⁰  -1.59876e-11 m²·⁰ s⁻¹·⁰
 -1.26268e-11 m²·⁰ s⁻¹·⁰   2.97524e-11 m²·⁰ s⁻¹·⁰  …   2.39754e-10 m²·⁰ s⁻¹·⁰   3.82218e-11 m²·⁰ s⁻¹·⁰
 -1.25549e-11 m²·⁰ s⁻¹·⁰   2.21644e-11 m²·⁰ s⁻¹·⁰      2.68475e-11 m²·⁰ s⁻¹·⁰   1.33622e-10 m²·⁰ s⁻¹·⁰
```

The tests are passing and I have updated the docs. Here is for the reproducibility with their paper:

Screenshot 2025-03-22 at 18 23 25

Screenshot 2025-03-22 at 17 19 07

Could someone check if the approach is GPU compatible? I believe so, but an expert point of view would be nice!

EDIT: I have also just noticed that the computation of the D matrix is not reverse AD compatible. It works fine with forward AD, and I don't think I can do much with the form of the equation I have to deal with. But feel free to correct me if I am wrong!


albert-de-montserrat (commented):
Just double checking, in [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/9df614633b786f13afe19c3acfb913e25f5d4db8/src/ChemicalDiffusion/ChemicalDiffusion.jl#L360) do you really intend to do the `exp` of a matrix, or broadcast `exp` over its elements? Also, should drop `@muladd` since there are no `fma`s in this expression.

Iddingsite:
> Just double checking, in [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/9df614633b786f13afe19c3acfb913e25f5d4db8/src/ChemicalDiffusion/ChemicalDiffusion.jl#L360) do you really intend to do the `exp` of a matrix, or broadcast `exp` over its elements? Also, should drop `@muladd` since there are no `fma`s in this expression.

No, I did not. Thx for the catch! Brings a nice speed up:

```julia
julia> @btime compute_D($melt_multicomponent, T = $(1500C));
  330.369 ns (0 allocations: 0 bytes)
```

compared to 1.600 μs before. Still not happy with the reverse AD but I think it is due to the Inverse of the matrix. Which I cannot do much about. 

And yeah, I think I should take 5 min one day to understand what `@muladd´ is actually doing....

albert-de-montserrat:
> > Just double checking, in [here](https://github.com/JuliaGeodynamics/GeoParams.jl/blob/9df614633b786f13afe19c3acfb913e25f5d4db8/src/ChemicalDiffusion/ChemicalDiffusion.jl#L360) do you really intend to do the `exp` of a matrix, or broadcast `exp` over its elements? Also, should drop `@muladd` since there are no `fma`s in this expression.
> 
> No, I did not. Thx for the catch! Brings a nice speed up:
> 
> ```julia
> julia> @btime compute_D($melt_multicomponent, T = $(1500C));
>   330.369 ns (0 allocations: 0 bytes)
> ```
> 
> compared to 1.600 μs before. It is still not happy with the reverse AD but I think it is due to the Inverse of the matrix. Which I cannot do much about.
> 
> And yeah, I think I should take 5 min one day to understand what `@muladd´ is actually doing....

Great! Because apart from the speed up, the `exp` of a matrix is not equivalent to the broadcasted `exp`.

`@muladd` checks an expression looking for chances to do `fma`s. Where `fma` stands for "fused multiply-add" and it's a hardware instruction that does `fma(x,y,z) = x * y + z` with only one floating point rounding, instead of two. Long story short, if in a expression there are no obvious/hidden `x * y + z`, there's no point on using this macro.

I haven't investigated the reverse AD issue, but if you suspect that the problem is `inv(w)`, could you try doing `(w\I)` instead? `I` is the identity matrix

Iddingsite:
> Great! Because apart from the speed up, the `exp` of a matrix is not equivalent to the broadcasted `exp`.

That's what I thought too, but it gives exactly the same result somehow. Maybe Julia is doing something internally. 

> `@muladd` checks an expression looking for chances to do `fma`s. Where `fma` stands for "fused multiply-add" and it's a hardware instruction that does `fma(x,y,z) = x * y + z` with only one floating point rounding, instead of two. Long story short, if in a expression there are no obvious/hidden `x * y + z`, there's no point on using this macro.

Thx! Finally a clear explanation on that topic haha
So I guess I can also remove it here:

```
    D = @muladd D0 * (((fO2 / dfO2)^nfO2)) * exp(aX * (X + bX)) * exp(-(Ea + (P - P0) * ΔV) / (R * T))
```

> I haven't investigated the reverse AD issue, but if you suspect that the problem is `inv(w)`, could you try doing `(w\I)` instead? `I` is the identity matrix

So, the StaticArray is not happy with that operation because it is not immutable (Even if I go from SVector to MVector). But for testing purposes, if I do:

```
@inline function compute_D(data::MeltMulticompDiffusionData; T = 1K, kwargs...)

    if T isa Quantity
        @unpack_units λD0, λEa, w, R = data

        # convert to K to prevent affine error with Celsius
        T = uconvert(K, T)
    else
        @unpack_val λD0, λEa, w, R = data
    end

    w2 = Array(w)

    # calculate diffusion matrix using eigenvalues and eigenvectors (see Eq. 4 in Guo and Zhang, 2020)
    D = w2 * (λD0 .* exp.(.- λEa ./ (R .* T))) * (w2 \ I)

    return D
end
```
that gives indeed the same result. But the reversed AD still doesn't work, so I think the inv(w) is not the issue. 

The temperature dependence is only in the central term, so I think `w` doesn't have an impact there. And indeed, if I simply remove it:

```
D = (λD0 .* exp.(.- λEa ./ (R .* T)))
```

I find the same value for forward AD, but the reverse AD still gets stuck. So somehow, it is coming from this term I think.

albert-de-montserrat:
Btw, in `D = w * λD0 * exp(-λEa / (R * T)) * inv(w)`, you can actually precompute `w * λD0` and `inv(w)`, and store them in the struct. Would speed things up.

Iddingsite:
The correct expression is `D = w * (λD0 .* exp.(.- λEa ./ (R * T))) * inv(w)`, so I don't think I can for `w * λD0`, but I can definitely do that for inv(w), yes.
That brings again a nice speedup:

```julia
@btime compute_D($melt_multicomponent, T = $(1500C));
133.513 ns (0 allocations: 0 bytes)
```
compared to 330.369 ns (0 allocations: 0 bytes) before!

Iddingsite:
Ok, so after discussing with @albert-de-montserrat, I think there is no possibility to make reverse AD works here without allocating. So it should be fine to ignore it here for now. So this is ready for review and to be merged on my side!

Issue #282, opened 2025-07-01
Type conversion with Julia 1.12


aelligp:
There's a type mismatch when running the CI tests of JustRelax using the pre-release of Julia on ubuntu and windows for the DislocationCreep:

```julia
ERROR: LoadError: MethodError: Cannot `convert` an object of type 
  GeoParams.Units.GeoUnit{Float32,Unitful.FreeUnits{(MPa^-7/2, s^-1),𝐋^7/2 𝐓^6 𝐌^-7/2,nothing}} to an object of type 
  GeoParams.Units.GeoUnit{Float64,Unitful.FreeUnits{(MPa^-7/2, s^-1),𝐋^7/2 𝐓^6 𝐌^-7/2,nothing}}
The function `convert` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  (::Type{GeoParams.Units.GeoUnit{T, U}} where {T, U})(::Any, ::Any, ::Any)
   @ GeoParams C:\Users\runneradmin\.julia\packages\GeoParams\IuoXh\src\Units.jl:124
  GeoParams.Units.GeoUnit{T, U}(::Any) where {T, U}
   @ GeoParams C:\Users\runneradmin\.julia\packages\GeoParams\IuoXh\src\Units.jl:152
  convert(::Type{GeoParams.Units.GeoUnit{T, U}}, ::Unitful.Quantity{T, U, X}) where {T, U, X}
   @ GeoParams C:\Users\runneradmin\.julia\packages\GeoParams\IuoXh\src\Units.jl:160
  ...
see [here for the entire stacktrace](https://github.com/PTsolvers/JustRelax.jl/actions/runs/15928136755/job/44930344223#step:6:884)

aelligp:
addressed in #285

PR #284, opened 2025-07-31
Fix precompilation of GP in Julia 1.12

aelligp:
Addresses #282


aelligp:
closed in favor of #285

PR #293, opened 2025-09-22
Add `dimensionalize_and_strip`

albert-de-montserrat:
Some times, for e.g. plotting, one needs to dimensionalize some variable and strip it from its units. Currently this is a bit verbose and annoying as one needs to do
```julia
ustrip(dimensionalize(var, units, char_lenghts))
```

This PR just adds a couple of shortcuts to make it look a bit nicer:
```julia
dimensionalize_and_strip(var, units, char_lenghts)
```
or
```julia
@dimstrip var units char_lenghts
```


boriskaus:
Nice - can you add info about this in the docs?

PR #319, opened 2026-03-26
[ADDITION] Tensile plasticity of Popov et al. 2025 - struct and yield surface

aelligp:
- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Other, please explain

This PR introduces the necessary structure and computational routines for the tensile plasticity of [Popov et al. 2025](https://gmd.copernicus.org/articles/18/7035/2025/).
also fixes a few bugs I encountered along the way. 

the AD tests in the plasticity dont work anymore since we switched to the plain strain 2nd invariant computation. Weren't picked up due to a typo in the routine. Also we never actually tested the DP_regularised functionalities. this is now fixed

**Checklist**
- [x] New documentation related to the new feature was added


albert-de-montserrat (commented):
Since there are different ways to cap DP, perhaps it's better to call it `DruckerPragerCapPopov`, or similar, so its a bit less generic.

boriskaus:
> > Since there are different ways to cap DP, perhaps it's better to call it `DruckerPragerCapPopov`, or similar, so its a bit less generic.
> 
> I thought about TensilePopov() but DPCap is the same as the RheologyCalculator struct
> 
> Other Tensile plasticity laws like Li et al could be then added as TensileLi()

I disagree that this should have the names of the authors in the plasticity model. DruckerPragerCap is fine, the one of Li et al. is a HyperbolicTensile formulation, so you could use that as a name.

Also, I am somewhat sceptical that this will work correct as-is, because it would require local iterations for tauII, P, and \lamba_dot, whereas I believe that the current version of GP only does it for tauII and \lambda_dot. 
So you can either 
1) rename this PR anchor clarify that you only implement the yield surface 
2) add correct local iterations

In any case, the Popov et al. paper has 0D benchmarks to exactly test this, which we reproduce in RC (see readme). This would have to be added as well test.

albert-de-montserrat (changes requested):

aelligp:
Plotting the yield surface and flow potential reproduces the plots from the paper:

image


GeoParams currently fails to reproduce the 0D tests with local iterations due to its design. So I will leave this part for later and/or for the implementation within RC (which was intended anyways). 

For an implementation point of view this PR is ready @albert-de-montserrat @boriskaus

aelligp:
@albert-de-montserrat StackOverflow error in the TP_Conductivity has been fixed. I added GMG to the downstream testing as it seems capture the overflow test as well. can be removed in case we want to avoid the failing test with gridap. 


EDIT: Error has been fixed by [#196](https://github.com/JuliaGeodynamics/GeophysicalModelGenerator.jl/actions/runs/23854021801/job/69541817156?pr=196)

albert-de-montserrat (approved):

PR #325, opened 2026-04-13
[UNITS] Move to SI units

albert-de-montserrat:
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Other, please explain

To avoid potential bugs, this PR moves all the default values and databases to SI units.

**Checklist**
- [x] New documentation related to the new feature was added


Iddingsite:
Is that really a good idea? I feel that being able to quickly check the implemented values with the values from papers is quite nice to check/compare the implementation with the literature. Especially for old papers that do not use SI.

boriskaus:
Can we not autoconvert all to SI when we specify them such that they are stored as SI but can be defined in say MPa?

PR #329, opened 2026-06-15
[FIX] A few bugs and ambiguities

aelligp:
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Other, please explain

While running JustRelax models in nondimensional form, @albert-de-montserrat  and I observed some weird cases where the plastic return mapping would not work although everything was properly nondimensionalised in the script. If the same script was run in dimensional form, the return mapping worked. 

This PR addresses these issues. Note that the src code audit and tests were performed and added by Claude Opus 4.8 and verified by me. 

**Checklist**
- [ ] The new feature was added in a way that does not break public API
- [ ] New documentation related to the new feature was added

### DIFF
--- a/src/ConstitutiveRelationships.jl
+++ b/src/ConstitutiveRelationships.jl
@@ -12,7 +12,7 @@ import GeoParams: second_invariant, second_invariant_staggered, value_and_partia
 using BibTeX
 using ..MaterialParameters: MaterialParamsInfo
 import Base.show
-using ForwardDiff, StaticArrays, Static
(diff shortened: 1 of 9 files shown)


albert-de-montserrat (approved):
So what triggered the errors in JR?

aelligp:
@albert-de-montserrat there was a hard coded `1` as the `hi` value in in the LinearSoftening. Also by not providing units in the softening, GP assumed nondimensionality and didnt scale the slopes or values correctly. I patched the bugs and a JR PR will follow that basically removed all `ustrip()` from the softening call.

PR #339, opened 2026-08-05
[ADDITION] Add standalone Unitful and DynamicQuantities backends

albert-de-montserrat:
### What's the purpose of this PR?

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Other: internal package reorganization

## Disclaimer

This is still in an early draft/prototype stage. Some design decisions may still change.

## Motivation

GeoParams' unit and nondimensionalization implementation previously lived in a single large
`src/Units.jl` file and could only be used by loading the full material-parameter package. This
made the unit layer harder to maintain, test, reuse, and evolve independently.

This PR follows the repository-local `lib/` package pattern to separate that functionality from
the GeoParams core. It also introduces a DynamicQuantities-backed implementation so applications
that need runtime dimensions can use the same GeoParams-specific units API without changing the
established Unitful behavior.

### Standalone unit packages

- Extracted the existing Unitful implementation into `lib/GeoParamsUnits`.
- Split the former monolithic implementation into focused files for constants, `GeoUnit`,
  characteristic scales, transformations, arithmetic, display, compatibility, and unpacking.
- Added `lib/GeoParamsDynamicUnits`, backed by DynamicQuantities.jl.
- Kept the two packages' explicit public export sets identical and verified numerical parity for
  representative transformations.
- Added package-local tests, Aqua checks, type-inference coverage, benchmarks, licenses, and
  READMEs.

### GeoParams integration

- Preserved `GeoParamsUnits` as the default backend and the historical `GeoParams.Units` path.
- Moved material-parameter traversal methods into `src/UnitsIntegration.jl`.
- Added `GeoParamsDynamicUnitsExt`, which activates when `GeoParamsDynamicUnits` is loaded.
- Added bidirectional conversions for quantities, `GeoUnit`, and `GeoUnits`, including arrays,
  fractional dimensions, custom geoscience units, and affine Celsius.
- Allowed DynamicQuantities inputs and characteristic scales at the GeoParams API boundary.

GeoParams material structures intentionally continue to store `GeoParamsUnits.GeoUnit`. This
preserves existing public types and serialized representations; the extension provides
interoperability rather than changing every internal field to a runtime-selected backend.

`using GeoParams` alone does not load DynamicQuantities or the dynamic extension.

### Documentation and CI

- Added a unit-backend guide and linked it from the manual navigation and homepage.
- Updated the root README, GeoUnit guide, and nondimensionalization guide.
- Added standalone README documentation for both `lib/` packages.
- Added independent CI matrices for both packages on Julia LTS, stable, and prerelease.
- Added an explicit GeoParams/DynamicQuantities extension-integration CI matrix.
- CI uses clean temporary environments so repository-local packages work across Julia versions.

## Compatibility

- The existing GeoParams public unit API remains available.
- `GeoParams.Units === GeoParamsUnits` remains true.
- The default Unitful-backed behavior is unchanged.
- `GeoParamsDynamicUnits` exposes the same public names and semantic behavior, although concrete
  quantity types necessarily come from DynamicQuantities.

## Validation

- `GeoParamsUnits`: 229/229 tests passing.
- `GeoParamsDynamicUnits`: 92/92 tests passing.
- Dynamic extension integration: 15/15 tests passing through `Pkg.test` on Julia 1.10.
- Existing root GeoUnits suite: 202/202 tests passing.
- Root Aqua suite on Julia 1.10: 53/53 tests passing.
- Both README quick-start examples execute successfully.
- The complete documentation site builds successfully.
- Runic formatting checks and `git diff --check` pass.

## Checklist

- [x] New tests were added, or old tests were updated.
- [x] Code coverage should remain the same or increase.
- [x] The new feature was added without breaking the existing public API.
- [x] Documentation related to the new feature was added.
- [x] The new code follows the [Runic Style](https://github.com/fredrikekre/Runic.jl).


albert-de-montserrat:
@Iddingsite

PR #344, opened 2026-09-02
[ADDITION] Evaluate material laws in the precision of their input - Float32 support

aelligp:
- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Other, please explain

Material laws evaluate in the precision of their *stored parameters*, which are
`Float64` throughout the shipped databases. A `Float32` stress handed to
`compute_εII` comes back as a `Float64`, so a GPU kernel with `Float32` arrays
cannot stay in `Float32` across a rheology call.

This PR makes a law evaluate in the precision of the state it is given.

**The mechanism** (`src/unpack.jl`): `precision_of(x)` reports the type a
calculation seeded by `x` should run in — a float selects its own type, while an
integer or a dual number expresses no preference and selects `Float64`, which
keeps constants off the AD tape. `convert_precision(T, x)` narrows a value while
retaining units. `@unpack_val T ρ, α = r` and `@unpack_units T ...` unpack
`GeoUnit` fields already converted, so a formula runs in the precision of its
solver state rather than that of its parameters. Stored parameters keep their
published `Float64` values; nothing is narrowed on disk.

**Scalar-or-array keyword arguments** (`argument_at` / `each_argument_index` in
`src/Utils.jl`): in-place routines required `T`, `P`, `f`, `d` to be arrays the
size of the destination, so a constant temperature had to be materialized as
`ones(size(TauII))`. They now accept either a scalar, shared by every element,
or an array indexed alongside the input.

**Range, and what it costs**

In SI units a dislocation law assembles an ordinary strain rate out of factors
that `Float32` cannot hold: prefactors reach 1e-55 and `(τ F_T)^n` reaches 1e39.
Six of the 42 laws overflow their intermediates in `Float32` at laboratory
stresses even though the result is an ordinary `Float32`. `retry_wider`
(`src/Utils.jl`) re-evaluates from a widened input when a result comes back
non-finite, or zero from a non-zero input, and narrows the answer back; all 42
laws then track their `Float64` reference to better than 1e-4 in both
directions. Two caveats worth knowing:

- Laws whose strain rate is *itself* outside `Float32` range (3e-71 for one)
  still flush to zero. That is not recoverable in `Float32`.
- The retry runs `Float64` arithmetic, which on a GPU is 1/32–1/64 throughput.
  It only fires for dimensional SI evaluation — the same clinopyroxene law
  nondimensionalized with `GEO_units()` never retries.

**Bugs fixed** (each verified against `main` in a worktree)

- `compute_param` dispatched back to itself when a parameter whose value takes
  arguments was called without any: `compute_density(PT_Density())` and
  `compute_shearheating(s, args)` both `StackOverflowError` on `main`. They are
  now `MethodError`s.
- The in-place elasticity routines left their own input array out of the index
  check, so a mismatched `τII` was silently truncated: a 5-element input into a
  4-element destination returned 4 values on `main`. Now `DimensionMismatch`.
- `ViscosityPartialMelt`'s in-place path dropped `ϕ`, disagreeing with its own
  scalar path (9.53e-5 vs 3.45e-6 on `main`). The expected value at
  `test_CreepLaw.jl:232` changes accordingly.
- `TASclassificationData` carried shared mutable `p`/`v` scratch fields, a data
  race across threads. They are now local to the call.

**Performance**

`@belapsed`, M-series CPU, `main` measured in a git worktree in the same
session. Run-to-run spread on the scalar calls is ~±1 ns, so differences below
that are noise.

| call | main | this PR |
|---|---|---|
| scalar, Float64 τ | 9.47 ns → `Float64` | 9.47 ns → `Float64` |
| scalar, Float32 τ | 9.47 ns → `Float64` | 9.50 ns → `Float32` |
| scalar, Float32 τ, law that overflows | 9.50 ns → `Float64` | 18.29 ns → `Float32` |
| `compute_εII!`, 1e6 elements, Float64 | 9.92 ms + ~23 MiB of kwarg arrays | 8.87 ms |
| `compute_εII!`, 1e6 elements, Float32 | 9.56 ms | 9.18 ms |

`Float64` is untouched: `_wider(Float64) === Float64` is a type-level
comparison, so the guard folds away at compile time and the emitted code is the
old expression. `Float32` costs nothing measurable until the retry fires, where
it is 1.9× — for a correct answer where `Float32` alone produces `Inf`. The
in-place path gains 11% in `Float64` by not indexing three keyword arrays.
Both scalar paths allocate nothing.

An earlier version folded `A^(1/n)` into the base of the power instead of
branching. It was algebraically cleaner but cost 2.13× on the `Float64` path, so
it was dropped.

**Behaviour changes a caller could notice**

- A law given `Float32` state returns `Float32` where it returned `Float64`.
  `Float64` results are bit-identical to `main`.
- `compute_density(PT_Density())` and `compute_shearheating(s, args)` raise
  `MethodError` instead of overflowing the stack.
- `compute_hb_viscosity_τII` raises on non-convergence rather than returning the
  100th iterate, and its tolerance is now `√eps` of the working precision.
- In-place routines accept scalar keyword arguments and reject mismatched arrays.

**Checklist**
- [ ] Are all lines covered by test? Code coverage should remain the same or increase
- [ ] New documentation related to the new feature was added

Tests: new `test/test_Float32.jl` (113 tests) and `test/test_Computations.jl`;
the whole dislocation database is exercised in `Float32` inside the existing
loop in `test_DislocationCreep.jl`; `DimensionMismatch` for all four in-place
elasticity routines; the shear-heating `MethodError`; an integer TAS
composition. All 41 test files pass locally, including Aqua, AD, allocations,
type stability and the headless CairoMakie tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/CreepLaw/MeltViscosity.jl
```diff
@@ -188,15 +204,17 @@ end
...
-    γ = 3.987815 * tanh(0.8908 * θ + 3.24) + 5.099645
+    ϕ_max = Tc(0.066499) * tanh(Tc(0.913424) * θ + Tc(3.850623)) + Tc(0.591806)
+    δ = Tc(-6.301095) * tanh(Tc(0.818496) * θ + Tc(2.86)) + Tc(7.462405)
+    α = Tc(-0.000378) * tanh(Tc(1.148101) * θ + Tc(3.92)) + Tc(0.999572)
+    γ = Tc(3.987815) * tanh(Tc(0.8908) * θ + Tc(3.24)) + Tc(5.099645)
     num = 1 + (c_vf / ϕ_max)^δ
-    x = √π * c_vf / (2 * α * ϕ_max) * (1 + (c_vf / ϕ_max)^γ)
-    den = (1 - α * erf(x))^(2.5 * ϕ_max)
+    x = sqrt(Tc(π)) * c_vf / (2 * α * ϕ_max) * (1 + (c_vf / ϕ_max)^γ)
+    den = (1 - α * erf(x))^(Tc(2.5) * ϕ_max)
```
albert-de-montserrat:
```suggestion
    den = (1 - α * erf(x))^(5 / 2 * ϕ_max)
```
Maybe?

albert-de-montserrat (commented):
I guess I'm fine with it, looks a bit uglier than before. But it is what it is :)

src/CreepLaw/MeltViscosity.jl
aelligp: its 5 // 2 now to preserve F32. 5 / 2 would promote it to F64 IIRC

